PR #2, opened 2016-03-17
unaligned_read_le32toh is makred as 64bit version.

eric-bentley:
The unalignedread_le32toh macro is incorrectly makred as
unalignedread_le64toh and the correct 64 bit version follows afterwards.

### DIFF
--- a/external/hash/unaligned.h
+++ b/external/hash/unaligned.h
@@ -21,7 +21,7 @@
         (((uint16_t)(((uint8_t *)(p))[0])) <<  0) |                         \
         (((uint16_t)(((uint8_t *)(p))[1])) <<  8))
 
-#define unaligned_read_le64toh(p)  (                                        \
+#define unaligned_read_le32toh(p)  (                                        \
         (((uint32_t)(((uint8_t *)(p))[0])) <<  0) |                         \
         (((uint32_t)(((uint8_t *)(p))[1])) <<  8) |                         \
         (((uint32_t)(((uint8_t *)(p))[2])) << 16) |                         \


mikkelfj:
The same error is found in https://github.com/dvidelabs/flatcc/blob/master/include/portable/punaligned.h which will be fixed separately. There has been a few other similar bugs already fixed, thanks for catching this.

eric-bentley:
Just glad I could contribute.  Likely would not have seen this except one end of our buffer usage is on an ARM processor.

I have a general flat buffer question for you:  We are using flat buffers to send responses of many different types to command we send over our ssh session.  I would like to have a union of off the different buffer types and then be able to do a switch() on the different buffer types in order to route to the appropriate handler for the current buffer type.  I did not see a way to do this other then if/then/else if/… through all the buffer types.  The alternative would be to append buffer type field before the actual buffer but I’d prefer not to do that if there’s a native way.

Did I miss something more appropriate?

Thanks - and enjoy the weekend - we have sun shine and warm for this first weekend in a long while here.

-Eric

mikkelfj:
Hi Eric, appreciated. Yes, as is stated in the README intro, the portability layer is not well tested for these special use cases, but it should be easy to get working once you start using it, it's not a lot of code.

The general question is probably best answered in the forum - but since you posted here: it is a very flatbuffer common use case you mention. You typically create a union of flatbuffers, like you mention. Creating an extra type field, as you suggest, is exactly what is done with unions under the covers. I'm not sure what you would if/then/else test on if you are not aware of this field, since table access would crash if done with the wrong type cast. Here is an if condition in the monster test, but it could as well have been a switch: https://github.com/dvidelabs/flatcc/blob/master/test/monster_test/monster_test.c#L454

If a union is named "message", the type field is named "message_type" and is an enum where each member is named after the table type. It is easy to see this enum in the generated code. The missing table in the union is 0 (NONE). There is a limit of 255 different types in a union.

We are also getting warmer weather here in Denmark :)

PR #3, opened 2016-04-28
build: Specify C language explicitly for CMake

derosier:
If the build system doesn't have a C++ compiler installed, CMake will fail
with a line like:
Check for working CXX compiler: CMAKE_CXX_COMPILER_FULLPATH-NOTFOUND

As flatcc is explicitly C-only, let's be sure we can use it without a C++
compiler installed.

Signed-off-by: Steve deRosier <steve.derosier@lairdtech.com>


mikkelfj:
Thanks!

PR #16, opened 2016-11-28
Fixes in JSON parser

skhoroshavin:
I've started playing with flatcc with simple schemes, and found that when there are no global enums defined generated json parser is uncompilable. Then, it was unable to parse input containing empty objects or arrays ( {}, [] ). Wasn't too hard to fix :)


mikkelfj:
Thanks a lot, that looks reasonable. Glad you found it easy to fix. I'll get back when I have verified it.

skhoroshavin:
Added fix for #17, wasn't easy as previous ones. Thanks a lot for your help!

PR #18, opened 2016-11-30
Added tests covering fixed bugs

skhoroshavin:
Added some tests to make sure that fixed bugs stay fixed :)

### DIFF
--- a/test/json_test/test_json.c
+++ b/test/json_test/test_json.c
@@ -329,6 +329,17 @@ int main()
             "{\"name\":\"Monster\",\"testnestedflatbuffer\":{\"name\":\"sub Monster\"}}");
 #endif
 
+    /* Test empty table */
+    TEST(   "{ name: \"Monster\", testempty: {} }",
+            "{\"name\":\"Monster\",\"testempty\":{}}" );
+
+    /* Test empty array */
+    TEST(   "{ name: \"Monster\", testarrayoftables: [] }",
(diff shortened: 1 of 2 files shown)


mikkelfj:
Thanks - I didn't dare ask. Could you please rename TestIssue17 to something more telling. The test schema might also be used in other contexts and is loosely based on the flatc test schema.
    
    table TestIssue17

=>

    table TestJSONPrefixParsing

skhoroshavin:
No problem, done.

mikkelfj:
note that I squashed

skhoroshavin:
Thanks. I was thinking about squashing commits myself and force pushing, but wasn't sure if this is handled correctly by github.

mikkelfj:
I added some more test cases to cover the exact key length of 8, which was the reason for the original problem:
https://github.com/dvidelabs/flatcc/commit/78f35eff734d293e2e2ef2475a4ddc95c65b9eb4

skhoroshavin:
Unfortunately there is a problem with your last commit - it no longer tests case when n != 8. You can check this by undoing my fix to parser generator and seeing that all tests pass. Or you can make a coverage report and see that this branch is no longer taken. Furthermore, I tried to undo your commit with additional test cases and checked coverage, and it shows that branch n == 8 is already taken by some other tests.

mikkelfj:
Yes, I started wondering if this was the case myself.

mikkelfj:
Reverted test case. BTW: github does support force commit but I used revert because this is on master branch

Issue #19, opened 2016-11-30
Add automated test coverage analysis


skhoroshavin:
I think that could help write tests that will cover more code, improving overall quality. I've made some quick hacks to generate report with gcov so you could see what it looks like, but in order to be efficient this better be automated and run regularly. If you agree to use gcov/lcov I could provide help, since I already had experience with these tools.
[flatcc-coverage.tar.gz](https://github.com/dvidelabs/flatcc/files/621638/flatcc-coverage.tar.gz)

mikkelfj:
I can see the idea, but in terms of cost benefit I'm not convinced about adding this to the general process. But you are welcome to use it and to add test cases where you find a need.

Some untested areas are really difficult or at least non-obvious and with low usage such as custom block alignment in the builder interface. Other more important tests would likely not be detected by gcov because they are concerned with exposing logic in the multiple layers of macro templating forming the user interface, and finally gcov would only expose obvious errors, not logical errors such as incorrect alignment relative to FB specs. So if one has to economise on the time spent on testing I would probably limit the gcov focus to the runtime libraries, and possibly exlcude the more esoteric interfaces until they are actually needed by someone.

There are also other tests that could need some attention. In the test directory there are some scripts test/leakcheck.sh etc. using valgrind that I used at some point - but got many false positives and now tend to rely on the compilers reporting, which isn't perfect, especially not without full coverage.

mikkelfj:
One test that would be good to have is one that exposes hash chain collisions - see recently fixed bug for issue #12.

mikkelfj:
BTW: I'm not familiar with gcov/lcov, but I don't mind adding it as an off by default build option  for gcc or clang compilers, and scripts could be added to tests/coverage/*.sh and possibly a driver in scripts/

skhoroshavin:
Well, I'm not advocating that coverage analysis is silver bullet which should be applied to all code nor that it actually finds errors. All it does is show which lines were executed and which branches were taken. My general workflow with this tool is the  following:
1) add/modify some code
2) run tests with coverage analysis
3) if tests fail fix the code by going to 1
4) if added code is already covered feel free to go to 1 for more hacking (or add simple test to show off new functionality)
5) cover code with additional tests or mark it as not needed to be covered (such as failing asserts etc)
6) go to 1

[...]

Issue #17, opened 2016-11-29
Corrupted JSON parser for some field name combinations


skhoroshavin:
Looks like JSON parser generator sometimes generates corrupted path for parsing _type fields. I've attached a simple repro case:
[test.fbs.txt](https://github.com/dvidelabs/flatcc/files/617888/test.fbs.txt)
[test.json.txt](https://github.com/dvidelabs/flatcc/files/617887/test.json.txt)

After lots of experiments I found that fields with length from 1 to 3, 8 to 11, etc work correctly, while lengths 4 to 7, 12 to 15 lead to nonparseable field_type. From comments in generated parser file it seems that some symbols get eaten. I tried to debug code generator, found that field dictionary is built correctly, and then something bad happens in gen_trie.

mikkelfj:
Thanks, I'll have a look. This is good info to go by. FYI: The trie is built in levels so 8 bytes are processed at a time. The exact lengths causing problems sound strange - it could also be the relative sorting order of keys being tested and the length of their shared prefixes.

skhoroshavin:
I've managed to reduce minimal test scheme to:

table test
{
	aaaa: string;
	aaaa12345: uint;
}
root_type test;

aaaa is read just fine, aaaa12345 is unrecognized. I think this confirms that problem is in trie generator, not unions. Relevant parts of generated code are following:

        buf = flatcc_json_parser_symbol_start(ctx, buf, end);
        w = flatcc_json_parser_symbol_part(buf, end);
        if (w == 0x6161616100000000) { /* prefix "aaaa" */
            buf += 8;
            w = flatcc_json_parser_symbol_part(buf, end);
            if ((w & 0xff00000000000000) == 0x3500000000000000) { /* "5" */
                buf = flatcc_json_parser_match_symbol(ctx, (mark = buf), end, 1);
                if (mark != buf) {

                    /* PARSE UINT */

                } else {
                    buf = flatcc_json_parser_unmatched_symbol(ctx, buf, end);
                }
            } else { /* "5" */
                buf = flatcc_json_parser_unmatched_symbol(ctx, buf, end);
            } /* "5" */
        } else { /* prefix "aaaa" */
            if ((w & 0xffffffff00000000) == 0x6161616100000000) { /* "aaaa" */
                buf = flatcc_json_parser_match_symbol(ctx, (mark = buf), end, 4);
                if (mark != buf) {

                    /* PARSE STRING */

                } else {
[...]

PR #20, opened 2016-11-30
Added coverage target to enable coverage testing with gcov/lcov

skhoroshavin:
Added option FLATCC_COVERAGE, which is OFF by default. It can be turned on only in debug build with tests enabled, and is useful only with gcc compiler. When turned on it enables generation of coverage information during build and run. It also adds "coverage" build target, which collects data saved during test runs and makes useful html reports. Related to #19

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,11 @@ option(FLATCC_INSTALL "enable build of runtime library only" OFF)
 # experience issues, and this option can then help.
 option(FLATCC_TEST "enable tests" ON)
 
+# Use with debug build with testing enabled only. Enables generation
+# of coverage information during build and run. Adds target "coverage"
+# which collects data and makes HTML report in build directory
+option(FLATCC_COVERAGE "enable coverage" OFF)
+
 # Affects the flatbuffer verify operation. Normally a verify should just
 # quickly reject invalid buffers but for troubleshooting, assertions can
 # enabled. This requires rebuilding the runtime library and will likely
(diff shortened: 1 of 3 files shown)


mikkelfj:
Thanks - but isn't --coverage compiler specific?

if (FLATCC_COVERAGE)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} --coverage -DNDEBUG")
endif()

Either one would then only enable if there is a supporting compiler, or the flag should only be set for for supported compilers can an error or warning printed. Suggestions?

mikkelfj:
I think you mentioned the use of a gcov or lcov library, but I only see a command line tool.

Wrt --output-directory, does this land in the build directory?. Otherwise it should, so it doesn't affect the git state and is easy to clean.

    COMMAND genhtml coverage.info --output-directory coverage)

mikkelfj:
I see you added reversal of #17 tests - this crossed my changes (and shouldn't really be in this PR anyway). Hence the branch conflict.

mikkelfj:
But your fixes are better than mine -

skhoroshavin:
gcov is a library that needs to be linked with when building with coverage information generation. It turned out that C flags are handed to linker too, and it's smart enough to link this library after getting option "--coverage"

lcov is a command-line tool to process raw data generated from gcov-enabled tests and output it into intermediate file. And genhtml is a command-line tool from lcov toolset to process data generated by lcov into nice html report.

> does this land in the build directory?
Yes, it does.

skhoroshavin:
As for tests - should I rebase my changes on your master and try again, or it's better to just revert them?

mikkelfj:
Rebase

mikkelfj:
I think coverage should be constrained to gcc - it does work with clang, but it is a second citizen
mind you - I haven't really worked with this
http://clang.llvm.org/docs/SourceBasedCodeCoverage.html

skhoroshavin:
> Thanks - but isn't --coverage compiler specific?

Yes, it's gcc specific, but since coverage option is off by default I don't think it's much of a problem - anyone enabling it explicitly will see comments and understand limitations. Supporting also clang-generated coverage is better, and would require checking compiler, but for now I'm inclined to leave this for somebody who is familiar with clang.

mikkelfj:
ok

PR #24, opened 2016-12-12
Fixed comments edge case in C reader code generation

skhoroshavin:
Fix was easy, but I still don't get how to properly test it. I've added repro case to cgen_test, but automatic tests still pass. I tried to feed output of cgen_test to gcc manually, and it doesn't compile until I apply my fix, so case is correct. It seems like automatic tests just don't check that it's actually compilable.


mikkelfj:
What if the documentation comment has the form?

/************ PRIMARY OBJECTIVE *************/

mikkelfj:
Regarding you comments to the cgen_test. It should check that it compiles. If it doesn't, something has gone wrong. Perhaps a return code in the transition from earler shell script driven testing towards CMake tests? It's actually important this works.

If one were to create solid comment testing, one would have to generate a very simple schema and match the generated output, similar to json tests, to ensure only doc comments, and all doc comments get through cleanly, using various styles that seem approapriate. I personally did not find this worthwhile doing, but did check things looked ok by manual inspection.

skhoroshavin:
> /************ PRIMARY OBJECTIVE *************/
This isn't recognized as doc, so it doesn't reach output.

mikkelfj:
To clarify, ideally
`/************ PRIMARY OBJECTIVE *************/`
should't become
`/** ********** PRIMARY OBJECTIVE *************/`
I'm sure there are other odd cases such as multi-line comments with "***" on each line that are not worthwhile handling, but the single line example is common enough so I suggest at least a test for that, something like
565: (untested)
```
fprintf(out->fp, "%s/**", indent);
/* Handles case where comment contains '////' or '****' */
if (doc->ident->len > 0 && doc->ident->text[0] != '*') {
    fprintf(out-fp, " ");
}
```

mikkelfj:
I don't recall the details for when doc is generated, and I trust you to do the sensible thing. The above was written before I read your answer.

skhoroshavin:
> It should check that it compiles. If it doesn't, something has gone wrong.

Double checked this, tests pass even with bug present. If it matters I'm on archlinux now, tomorrow I'll be able to check this on ubuntu 14.04 to see if it's system dependant.

mikkelfj:
It matters a lot. It would be great if you would have a look. This is the first line of defense for the parser and scope handling. I might look awfully messy, but for example, the `island` is used in different scopes and many odd edge cases are covered here.

skhoroshavin:
I'll check this. It would be very helpful if you actually pointed me at where its output is actually fed to compiler.

As for handling //// and **** comments - your code looks sensible enough, the only problem is that it seems like not every input comment is regarded as doc token, I tried to trace this back to lexer and then stuck in luther internals. Any guidance would be helpful.

mikkelfj:
As to Lex Luthor, he generates an array of tokens and pointers into the original buffer. Comments have their own type, and I don't recall details, but I believe there are single comment and block comment styles, but other than that buffer content provides clues (e.g. # vs //). The doc pointer might point a little further into the token to skip the comment symbol itself, but I don't recall, it just makes sense. The primary criteria for doc comments is what Wouter writes in the schema doc, and my choice to only accept comments before a type name since anything else would be messy. I'm not sure how blank lines are handled, but multiple comments (block or single) are added to the doc list of the type it is associated with (initially doc lists are stored separately, but are then transferred to the type object, or dropped if some other token arrives - this happens by assigning null to the parsers doc list somewhere at the end of parser.c. But this is only a rough outline from memory, and me reading the parser.c code quickly.

skhoroshavin:
Tried your code and bug is back. After adding some printfs to codegen
```
        fprintf(stderr, "linenum %ld, ln %ld, first %d, len %ld\n", doc->ident->linenum, ln, first, doc->ident->len );
        fprintf(stderr, "%.*s\n", (int)doc->ident->len, doc->ident->text);
```
I've got the following output:
```
...
linenum 9, ln 0, first 1, len 0

linenum 9, ln 9, first 0, len 55
///////////////////////////////////////////////////////
linenum 10, ln 9, first 0, len 0

linenum 10, ln 10, first 0, len 55
///////////////////////////////////////////////////////
...
```

mikkelfj:
The cgen test was originally placed here:
https://github.com/dvidelabs/flatcc/blob/master/test/cgen_test/cgen_test.sh

but is now placed here
https://github.com/dvidelabs/flatcc/blob/master/test/cgen_test/CMakeLists.txt

and I think it is obvious why it isn't working as intended.
A dependent build step is needed. I don't recall exactly what happened, this might not be just an oversight though it likely is.

Background: I couldn't get CMake to build dependencies of custom targets, so flatcc didn't build before custom targets that depended on it. This is why all the performance benchmarks are no in CMake (although they are in the experiment meson build on a separate branch).

mikkelfj:
On the Luthor thing, it could also be that comment token excludes the delimiter but that they are present in the buffer at offset -1 if one need to distinguish types. It seems to be the case from your output.

skhoroshavin:
> I couldn't get CMake to build dependencies of custom targets, so flatcc didn't build before custom targets that depended on it. 

I also struggled a lot with this problem, but finally got this working through add_dependency. You may look at [this](https://github.com/skhoroshavin/dsim/blob/master/CMakeLists.txt) quite dirty, but working example.

mikkelfj:
Luthor emits comments and event control codes, so a single comment may be quite fragmented. The parser interface to luthor is responsible for pushing the events onto the array and does some translation here.
https://github.com/dvidelabs/flatcc/blob/master/external/lex/tokens.h#L365

https://github.com/dvidelabs/flatcc/blob/master/src/compiler/parser.c#L1172-L1185

skhoroshavin:
> On the Luthor thing, it could also be that comment token excludes the delimiter but that they are present in the buffer at offset -1 if one need to distinguish types. It seems to be the case from your output.

Well, this happens with all comments in test case. Not sure if this is correct behavior and "first" check should be modified to be actually useful, or something should be done with parser itself.

mikkelfj:
> I also struggled a lot with this problem, but finally got this working through add_dependency. You may look at this quite dirty, but working example.

I looked at all sorts of examples and tried lots of combinations, and eventually I just didn't want to spend more time on it.

In comparison meson dependencies are tight as a drum, especially with flatcc dependency file generation, despite the tool being a bit rough around custom builds. You can update a comment in a second level included schema and it will rebuild the whole project in a few milliseconds.

mikkelfj:
> Well, this happens with all comments in test case. Not sure if this is correct behavior and "first" check should be modified to be actually useful, or something should be done with parser itself.

Don't go overboard in this - I think first is just the head of a collection of comments that weren't interrupted by something else. That something else is likely blank space that is learned by detecting holes in line number sequences. It seemed to work at the time, and likely still does. But if is a headache, just add the space and let it be until one day it annoys you enough to fix proper.

mikkelfj:
Just to be clear: the luthor comment begin event contains /* or //. This is followed by one or more (possibly empty) comment body tokens followed by a comment end token (on valid parse). Only the comment body is pushed onto the parsers token stack and this is why the lead symbols appear to be missing. But they are guaranteed present at negative offset if needed. Control characters may also be emitted, but mostly for error reporting and symmetry with string handling.

Also, luthor has an option to detect doc comments, i.e. /** style comments. But I don't recall how this works here, only that the is_doc flag is being forwarded from luthor to parser.c and that this does not necessarily directly relate the flatbuffer doc concept, but it might.

skhoroshavin:
Okay, I'm starting to understand, but it's 4 AM, so I need to get some sleep ) Hope to fix this tomorrow.

skhoroshavin:
Updated PR with changes to CMakeList.txt so that generated source is actually compiled during testing. Not ideal but working for me. I'm also going to look more closely into parser later.

UPD. Very strange, CI shows tests failing, I'll try to fix this.

skhoroshavin:
Seems like tests working correctly with make, but always fail with ninja. Tried to make a quick fix but failed again. Any ideas?

mikkelfj:
> Seems like tests working correctly with make, but always fail with ninja. Tried to make a quick fix but failed again. Any ideas?

Yes, dependencies doesn't work in CMake for custom targets.

skhoroshavin:
> Yes, dependencies doesn't work in CMake for custom targets.

Actually, they work when using make build system, but as far as I can see they don't work for anything else (msbuild & ninja)

mikkelfj:
The problem is also present in MSVC builds:
https://ci.appveyor.com/project/dvidelabs/flatcc/build/win%23292/job/4d2er9xouoq85df3
line 3114
```1: 
 1/16 Test  #1: cgen_test ........................   Passed    0.02 sec
test 2
      Start  2: test_generated
2: Test command: "C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe" "test_generated"
2: Test timeout computed to be: 1500
2: Microsoft (R) Build Engine version 14.0.25420.1
2: Copyright (C) Microsoft Corporation. All rights reserved.
2: 
2: MSBUILD : error MSB1009: Project file does not exist.
2: Switch: test_generated
 2/16 Test  #2: test_generated ...................***Failed    0.08 sec
test 3
      Start  3: monster_test
```

mikkelfj:
> Actually, they work when using make build system, but as far as I can see they don't work for anything else (msbuild & ninja)

No necessarily, it might also fail with concurrent make builds.

skhoroshavin:
> No necessarily, it might also fail with concurrent make builds.

Never seen this. And we are using concurrent make builds heavily on my primary job. I've made a quick hack so that test_generated is run only for make build system. Not good, but better than nothing.

mikkelfj:
Just add a comment along the lines:

Compilation of the generated code tests many import edge cases in the parser and code generator but due to CMake limitations, custom target dependencies only work for Make build targets.

mikkelfj:
BTW: upgrading CMake won't work because Ubuntu distributions don't have very recent CMake.

skhoroshavin:
> Just add a comment along the lines:

Done

mikkelfj:
Thanks, should we close it here, or do you wish to explore further?

skhoroshavin:
I'm going to explore this further, but I'd prefer not to block this fix, since this bug was quite annoying. What about closing this and opening new low priority issue "Improve comments formatting"?

mikkelfj:
Or just close the issue, and then you can later create a PR if you feel like it.

Issue #23, opened 2016-12-05
Add "scan_by" methods to reader interface


skhoroshavin:
Sometimes order of items in array is important, yet it there could be need to find some item in array by some key. In this case the only option is to perform linear search, which is still okay for small number of items (and small number of items seem to be quite frequent case for flatbuffers). Writing search functions manually can be tedious, so it could be great if flatcc generated them automatically, either giving them suffix _unsorted, or generating them in place of usual binary searches when some key is given to flatcc. Or maybe some other way (ukey attribute, instead of key?)

mikkelfj:
Flatbuffers schema has a `key` attribute but it is only allowed on one table field. The flatcc schema parser actually allows for multiple key fields and generates `find_by_`... accessors, as well as just `find` for the first key.

The builder does no sort arrays because of the emitter interface that supports custom streaming and avoids storing the full buffer in memory at once. Instead the emitter interface has a cast to mutable and a sort operation on mutable arrays. There is a sort for each keyed field. The sort is an in-place unstable heap sort that does not require extra memory. The is documented in the builder.md document and the reflection api (bfbs) implementation makes use of this because binary flatbuffers have sorted arrays.

In addition it is possible to manually sort arrays by other criteria and implement own binary search or hash lookup methods - but this is beyond the scope of the core flatcc implementation. That said, support libraries could be developed for this purpose.

Note that sorting is unsafe even with a verified buffer - you need to be able to trust the buffer 100% or to first rewrite the buffer due to certain attack vectors related to overlapping fields.

mikkelfj:
Giving it some more thought, it would be possible to add a `scan_by` method to the reader interface. This could be done on all keyed fields. It would be logical when supporting multiple keys, since at most only one of these would be sorted.

While the binary search is non-trivial due to the macro system, a scan should be able to reuse the comparison logic and therefore be relatively straight forward to implement.

If you are up for it, without any modifications to the schema, please have a go.

skhoroshavin:
Thanks, I know about sort helper functions, I already read all available documentation on both flatbuffers and flatcc project. My main concern was not about arrays being unsorted by default, but about _keeping_ them unsorted, and still being able to find some element by key. For example, order of stored fields of some structure definition is important, and yet there could be need to search for some field by name. However, on the second thought in this specific case fields will likely have something like offset property, and can be sorted back into original order when needed. And even if they don't have one, it can be easily added. So, having sort and binary find functions for different fields is flexible enough for any task.

[...]

PR #25, opened 2016-12-13
scan_by implementation

skhoroshavin:
Core functionality (scan_by/scan_at_by/scan_range_by) discussed in #23 implemented and tested, including scalar fields.


mikkelfj:
This is a mostly complete implementation, but scan_range needs test, some refactoring is required to contain search logic in a macro, and doc needs to be added. But usable as is.

PR #26, opened 2016-12-14
Various improvements to scan functionality and tests

skhoroshavin:
- wrapped generation of find/scan function in macro
- added tests for scan_range
- added scan_next and tests for it

Unfortunately, it's not possible to wrap defines for default keys in macro as is (at least I don't know any way to add define into define), I see here two possibilities:
1. Make wrapper function which generate all those defines to reduce duplication in codegen
2. Switch from plain defines to inline functions so that they can be enclosed in macro

Also it might be a good idea to merge gen_find and gen_scan into one bigger gen_search, and merge various define_find/scan* into corresponding define_search*.

### DIFF
--- a/src/compiler/codegen_c_reader.c
+++ b/src/compiler/codegen_c_reader.c
@@ -80,8 +80,9 @@ static void gen_find(fb_output_t *out)
     fprintf(out->fp,
         "#include <string.h>\n"
         "static size_t %snot_found = (size_t)-1;\n"
-        "#define __%sidentity(n) (n)\n",
-        nsc, nsc);
+        "#define __%sidentity(n) (n)\n"
+        "#define __%smin(a, b) ((a) < (b) ? (a) : (b))\n",
+        nsc, nsc, nsc);
     fprintf(out->fp,
         "/* Subtraction doesn't work for unsigned types. */\n"
         "#define __%sscalar_cmp(x, y, n) ((x) < (y) ? -1 : (x) > (y))\n"
@@ -134,6 +135,26 @@ static void gen_find(fb_output_t *out)
         "static inline size_t N ## _vec_find(N ## _vec_t vec, T key)\\\n"
         "__%sfind_by_scalar_field(__%sidentity, vec, N ## _vec_at, N ## _vec_len, key, T)\n",
         nsc, nsc, nsc);
+    fprintf(out->fp,
+        "#define __%sdefine_find_by_string_field(N, NK) \\\n"
+        "/* Note: find only works on vectors sorted by this field. */\\\n"
+        "static inline size_t N ## _vec_find_by_ ## NK(N ## _vec_t vec, const char *s)\\\n"
+        "__%sfind_by_string_field(N ## _ ## NK, vec, N ## _vec_at, N ## _vec_len, s)\\\n"
+        "static inline size_t N ## _vec_find_n_by_ ## NK(N ## _vec_t vec, const char *s, int n)\\\n"
+        "__%sfind_by_string_n_field(N ## _ ## NK, vec, N ## _vec_at, N ## _vec_len, s, n)\n",
+        nsc, nsc, nsc);
+    fprintf(out->fp,
+        "#define __%sdefine_default_find_by_scalar_field(N, NK, TK)\\\n"
+        "static inline size_t N ## _vec_find(N ## _vec_t vec, TK key)\\\n"
+        "{ return N ## _vec_find_by_ ## NK(vec, key); }\n",
+        nsc);
+    fprintf(out->fp,
+        "#define __%sdefine_default_find_by_string_field(N, NK) \\\n"
+        "static inline size_t N ## _vec_find(N ## _vec_t vec, const char *s)\\\n"
+        "{ return N ## _vec_find_by_ ## NK(vec, s); }\\\n"
+        "static inline size_t N ## _vec_find_n(N ## _vec_t vec, const char *s, int n)\\\n"
+        "{ return N ## _vec_find_n_by_ ## NK(vec, s, n); }\n",
+        nsc);
 }
 
 /* Linearly finds first occurrence of matching key, doesn't require vector to be sorted. */
@@ -177,13 +198,10 @@ static void gen_scan(fb_output_t *out)
         nsc, nsc, nsc, nsc, nsc,
         nsc, nsc, nsc, nsc, nsc, nsc);
     fprintf(out->fp,
-        "#define __%smin(a, b) ((a) < (b) ? (a) : (b))\n",
-        nsc );
-    fprintf(out->fp,
         "#define __%sdefine_scan_by_scalar_field(N, NK, TK)\\\n"
         "static inline size_t N ## _vec_scan_by_ ## NK(N ## _vec_t vec, TK key)\\\n"
         "__%sscan_by_scalar_field(0, N ## _vec_len(vec), N ## _ ## NK, vec, N ## _vec_at, N ## _vec_len, key, TK)\\\n"
-        "static inline size_t N ## _vec_scan_at_by_ ## NK(N ## _vec_t vec, size_t pos, TK key)\\\n"
+        "static inline size_t N ## _vec_scan_from_by_ ## NK(N ## _vec_t vec, size_t pos, TK key)\\\n"
         "__%sscan_by_scalar_field(pos, N ## _vec_len(vec), N ## _ ## NK, vec, N ## _vec_at, N ## _vec_len, key, TK)\\\n"
         "static inline size_t N ## _vec_scan_range_by_ ## NK(N ## _vec_t vec, size_t begin, size_t end, TK key)\\\n"
         "__%sscan_by_scalar_field(begin, __%smin(end, N ## _vec_len(vec)), N ## _ ## NK, vec, N ## _vec_at, N ## _vec_len, key, TK)\n",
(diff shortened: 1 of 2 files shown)


mikkelfj:
Good job.

I had the same thought about gen_find, but it is isolated to the code generator internals it is already contained in gen_helpers, so I think it is ok as is.

Major thing: it seems that only string search operations are handled. There can also be scalar keys. Possibly even enums, and struct vectors keyed by member field. These last variants might be non-standard, but you can see where `find` is implemented and key attribute tests placed in code gen.

I think we need inlines for default key - I have had the same considerations earlier for other operations, it just wasn't worthwhile with a single find.

From the following example, I am not convinced that scan_next buys anything? If so I'd rather just stick with _at - not that _next wouldn't often be more useful, but _at would still be required. The justificication would be a typical example where you do not test for `not_found` before the next search.

```
static int count_monsters(ns(Monster_vec_t) monsters, const char *name)
 {
     size_t i;
     int count = 0;
 
     for (i = ns(Monster_vec_scan)(monsters, name);
          i != nsc(not_found);
          i = ns(Monster_vec_scan_next)(monsters, i, name)) {
         ++count;
     }
 
     return count;
 }
```

```
static int count_monsters(ns(Monster_vec_t) monsters, const char *name)
 {
     size_t i, count;
 
     for (i = ns(Monster_vec_scan_at)(monsters, name, 0);
              i != nsc(not_found);
              i = ns(Monster_vec_scan_at)(monsters, i + 1, name)) {
         ++count;
     }
 
     return count;
 }
```

```
static int count_monsters(ns(Monster_vec_t) monsters, const char *name)
 {
     size_t i, count, k;

     k = ns(Monster_vec_scan)(monsters, "Joker");
     for (i = ns(Monster_vec_scan)(monsters, name);
              i < k;
              i = ns(Monster_vec_scan_range)(monsters, i + 1, k, name)) {
         ++count;
     }
 
     return count;
 }
```

Note that _at is more consistent with range, and _at is self-contained as first and second search operation. Next does not appear to to simplify the logic.

I think we need some more examples.

On coding style, multi-line typically uses double ident, but pragmatic to circumstances and code alredy present (especially pragmatic in compact defines in generated code). In scan_next impl: pos+1 breaks with style of space around binary operators elsewhere (not saying one is better, that is just the convention used).

`count` should probably be `size_t` not that I prefer it, but since counts are often used as together with indexes, and you get signed/unsigned warnings, and it takes away uncertainty about overflow.

mikkelfj:
An here the "Joker" is a lower bound, so we can use _at similar to range with "Joker" as upper bound.

static int count_monsters(ns(Monster_vec_t) monsters, const char *name)
 {
     size_t i, count, k;
 
     k = ns(Monster_vec_scan)(monsters, "Joker");
     for (i = k; i != nsc(not_found);
              i = ns(Monster_vec_scan_at)(monsters, i + 1, name)) {
         ++count;
     }
 
     return count;
 }

mikkelfj:
More generally, on the potential TODO list:

It would be good with code generation to sort all vectors recursively by default key, similar to verify traversal. I just didn't find it worthwhile to implement, and am not sure that I will, so hesitate to create an issue. But if you find it valuable, it is up for grabs. This would avoid writing the following code manually, though it is simple enough to do:

https://github.com/dvidelabs/flatcc/blob/master/src/compiler/codegen_schema.c#L301

skhoroshavin:
> I had the same thought about gen_find, but it is isolated to the code generator internals it is already contained in gen_helpers, so I think it is ok as is.

I'll make a refactoring to show benefit.

> Major thing: it seems that only string search operations are handled. There can also be scalar keys. Possibly even enums, and struct vectors keyed by member field. These last variants might be non-standard, but you can see where find is implemented and key attribute tests placed in code gen.

Scalar field searches are implemented, they were already contained in nice macros and I just expanded them. Although adding some tests would make this more clear.

> I think we need inlines for default key - I have had the same considerations earlier for other operations, it just wasn't worthwhile with a single find.

Okay, on my todo.

> Note that _at is more consistent with _range, and _at is self-contained as first and second search operation. _next does not appear to simplify the logic.

Agreed. I'll remove _next wrappers. The only thing that really bugs me now is _at naming, I think _from is also short enough and shows intent more clearly.

> On coding style, multi-line typically uses double indent, but pragmatic to circumstances and code alredy present (especially pragmatic in compact defines in generated code). In scan_next impl: pos+1 breaks with style of space around binary operators elsewhere (not saying one is better, that is just the convention used).

I'm trying my best to adhere to what I see in a code, but sometimes habits from own projects are taking over. Anyway, thanks for pointing.

> count should probably be size_t not that I prefer it, but since counts are often used as together with indexes you void signed/unsigned warnings and potential overflows.

Okay.

> It would be good with code generation to sort all vectors recursively by default key, similar to verify traversal. I just didn't find it worthwhile to implement, and am not sure that I will, so hesitate to create an issue. But if you find it valuable, it is up for grabs.

On my todo, but not high priority.

mikkelfj:
Agree with all, include `_from`. You do a very good job with coding style in general. And I am not a fanatic, but it makes life simpler in the end to follow some pattern.

mikkelfj:
> Scalar field searches are implemented, they were already contained in nice macros and I just expanded them. Although adding some tests would make this more clear.

Hmm, it doesn't look like they are fully covered by macro wrapping:

https://github.com/skhoroshavin/flatcc/blob/591532f809463a9743d47ea8cdd99ed56b97759d/src/compiler/codegen_c_reader.c#L925-L929

https://github.com/skhoroshavin/flatcc/blob/591532f809463a9743d47ea8cdd99ed56b97759d/src/compiler/codegen_c_reader.c#L1227-L1231

skhoroshavin:
I meant these macros:
https://github.com/dvidelabs/flatcc/blob/master/src/compiler/codegen_c_reader.c#L127-L131
https://github.com/dvidelabs/flatcc/blob/master/src/compiler/codegen_c_reader.c#L182-L190

mikkelfj:
Sure, but you still end up with a gazillion defines for scalar vectors, even if they now go away for strings.

mikkelfj:
Sorry, the defines I listed where the ones blocked by not being inline functions, pending future work.

skhoroshavin:
I've updated PR with unification find/scan refactoring I was talking about yesterday so you can review it. If approved I'll continue other improvements (like converting some defines to inlines, removing _next, etc) on top of that.

mikkelfj:
It looks fine, but what if we decide to support scan on all fields, and find only for fields with keys?

skhoroshavin:
Good point. On the other hand scanning on all fields will generate tons of code, do we really need this?

skhoroshavin:
Btw, while I was waiting for review I've looked into that JSON parser bug I was talking recently. I didn't minimize repro case, but from what I seen by running under gdb it seems like a problem of handling unexpected fields in union type.

mikkelfj:
> Good point. On the other hand scanning on all fields will generate tons of code, do we really need this?

I wouldn't put scan on all fields by default, but it could be flagged. I'm not sure how much code it generates in terms of compilation overhead, but in terms of line count it would be zero, except for the common file, because it would be a standard field operation. But more generally, if there isn't significant benefit to this refactoring, also considering _at, _len are separate anyway, ... I have no strong feelings about this though.

skhoroshavin:
Okay, I'll rollback then.

mikkelfj:
> Btw, while I was waiting for review I've looked into that JSON parser bug I was talking recently. I didn't minimize repro case, but from what I seen by running under gdb it seems like a problem of handling unexpected fields in union type.

That is very likely. The union handling is very difficult and there has been similar bugs during development. It manages a complex set of states to ensure type field matches object field. It sometimes backtracks because the type field is late, and must manage this hierarchically on a stack. Finally, it must check that if a type was specified and it was not NONE, the object must be present and vice versa.

mikkelfj:
BTW: I don't think there is anything wrong in a high level search macro so there is only one search macro in generated code. I think that is desirable, I just don't think it needs to go all the way down. I haven't considered in detail how your solution deals with that, but perhaps you could keep some of it.

mikkelfj:
That is, if we choose to place scan everywhere, it would be flag that either enables it in search or in general vector ops. Not that we should, but we should allow for it, I think.

skhoroshavin:
For now I've split everything back into find/scan. I'll make other refactorings on top of that, and if/when we get a more concrete solution on unification I'll merge them back. It's really not a big deal now.

mikkelfj:
OK, sounds good.

skhoroshavin:
As for unions bug - I think I've catched it. Usually when there is unexpected symbol parser sets an error and then just unrolls to return as fast as possible. However, when this unexpected symbol occurs in union it starts to unroll, and then tries to parse again, triggering assert. Before patching this I think JSON testing framework should be extended to be able to expect parsing errors, if you don't mind.

mikkelfj:
It would be good with tests for JSON error cases.

As to unions, bail-out needs to be considered carefully. I think you can just return without dealing with the stack because there is at most one error and the builder is just left in bad state that must be reset. But still, the linenumber of the error is non-trivial because of the back-tracking. It resets the line counter when it does a second typed parse. This may still be trivial to fix, just noting potential issues.

skhoroshavin:
Ready to merge. Refactored:
- switched default field find/scan to inlines
- renamed scan_at to scan_from
- small fixes in tests

skhoroshavin:
Oh, again double indents in defines. My IDE must be kidding me. I'll fix that in a couple of minutes

mikkelfj:
`scan_n_range` is there any precedence for where to placing the `_n`?

I'm not sure which is best - `find_n_by` does not give any answer. My thinking is `_n` is associated with the key argument and should thus be placed late. `string_str_n` operations also has `n` late in the builder api.

coding style: I believe you sneaked in a space `nsc, nsc );` :)
count still uses int in test.

mikkelfj:
> Oh, again double indents in defines. My IDE must be kidding me. I'll fix that in a couple of minutes
pragmatic, sometimes it takes up too much space.

skhoroshavin:
> My thinking is _n is associated with the key argument and should thus be placed late

While scan_range_n_by_name is good scan_from_n_by_name could be a bit misleading. But I can rename them if really needed.

> coding style: I believe you sneaked in a space nsc, nsc ); :)

I'd better have some sleep :)

> count still uses int in test

[Really](https://github.com/dvidelabs/flatcc/pull/26/commits/d3b7a4ffc85480ecf0674da29252f9177b610e7d#diff-c2ebe16717f17f52e0f3fa858f005325R1105)? :)

mikkelfj:
> Really? :)

No, I just deleted that comment 2ms before.

skhoroshavin:
Removed extra space

skhoroshavin:
Preparing next PR:
1) should I rename scan_n_from_by_name to scan_from_n_by_name?
2) I'm starting to think that providing scan_by_name for every field really makes sense, maybe I could implement it by default?
3) following 2 - does it really makes sense to provide defaults for scan functions?

mikkelfj:
1. You sort of convinced me to keep it as is, but as I said, I am in doubt - no matter what you do it isn't ideal. We should settle before release, but keep it for now and give it some more thougt.
2. Ideally we should measure the overhead, but that depends on compiler. I think it should be flagged, but could be enabled by default. Say there is some old embedded systems compiler that has a hard time with many inline functions, then it can be disabled and revert to current.
3. yes, because scan should be equivalent to find. Not that many would use it, except if you sort by different criteria, but for consistency it should.

mikkelfj:
4. we also need from, range for find.
5. a find that finds lower lesser key if missing, or end of range if there are no lesser keys, but never not_found. It could have pointer arg to indicate match: lt<0, eq = 0, gt > 0, where gt is also returned on empty sequence.

On 2, I'm not sure how many would actually use it. scan can quickly become an expensive operation - I doubt I would use it a lot, unless it was to traverse sorted subranges, and the 5 is more useful if combined with a range.

skhoroshavin:
1) both variants are not ideal, but now I lean more to scan_range_n variant
2) just one example - in my project I have some sort of JSON-based data definition language, and when declaring structs I need both to leave their fields in original order, and to search by field name. And declaring field name as key implicitly means that fields should be sorted, which is incorrect. 
3) okay, I'll leave it as is
4) not sure about this. From/range are mainly for incremental searches, but for sorted ranges problem is often solved much better by lower/upper_bound or equal_range equivalent
5) actually I'd better implement equal range as base algorithm, and then make find/lower_bound/upper_bound a simple inline wrappers on top of that

Considering JSON-parser bug with unions - I've extended json_test, but still can't catch it in small repro case. However I could still make a small PR with new tests if you don't mind:
https://github.com/skhoroshavin/flatcc/commit/ad72c3b8ca6a36a6676c81a651d74e170742fceb

mikkelfj:
1. Change it if/when you are confident enough to make that decision.

2. yes, but you don't need large datasets performance goes way down - I'd soon start to add an external index. Just iterating 100 elements every incoming request can reduce throughput by a factor 10, depending on what you do, of course. For CSV header file identification, performance is completely irrelevant. So question is just, is is relevant enough to add as default.

4. For consistency, and because you might want to do incremental processing of a large dataset:  For example, if you have established an area code range, you only want to search within that range for more specific area codes. Or you may want to do an optimistic search (simplified exponential search), avoiding hitting all over the file if you expect the next match to be near the last one. So you search only the next 100 elements before searching 10000, then all. Avoids reading random data. Or you might have indexed the data elsewhere in intervals, say IP block ranges, and want to search for specific IP's in that range, or nearest lower in that range.

5. can be trivially and efficiently implemented on binary search whereas equal_range would need a scan or two searches. It seems equal range would be better implemented with an lbound / ubound search, where that last search is bounded by the first.

But we don't have to address that now.

mikkelfj:
> Considering JSON-parser bug with unions - I've extended json_test, but still can't catch it in small repro case. However I could still make a small PR with new tests if you don't mind:
skhoroshavin@ad72c3b

I think expect_err should be expect_err_line and verify the line of failure. For compact tests it will be just 1, but it is a current weakness that line numbering is not tested, especially for unions.

PR #27, opened 2016-12-16
Quick fix for rare JSON parser cras when unexpected field encountered

skhoroshavin:
Sometimes when failing to parse table parser crashes in flatcc_builder_end_table (failing check for some frame), so I added a quick fix to early out in this case. This definitely should be investigated further, but for now I'm out of ideas, and somewhat short of time. 

As for your comment on checking for error_line instead of expected_error - I think that both checks could be useful. Anyway, if you don't like current implementation you can just drop testing commit and accept fix only.

### DIFF
--- a/src/compiler/codegen_c_json_parser.c
+++ b/src/compiler/codegen_c_json_parser.c
@@ -562,6 +562,7 @@ repeat_nested:
         println(out, "buf = flatcc_json_parser_string_end(ctx, buf, end);");
     } else if (is_table) {
         println(out, "buf = %s_parse_json_table(ctx, buf, end);", snref.text);
+        println(out, "if (ctx->error) return buf; /* Quick fix, should be investigated further */");
         println(out, "ref = flatcc_builder_end_table(ctx->ctx);");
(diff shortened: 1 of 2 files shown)


mikkelfj:
I didn't actually mean instead of, I'd like both. But what you have is fine, just stating where there is a weak spot in testing.

I can't commit to debugging the parsing right now, but if you have a reproducible case, I'll get on it soon.

mikkelfj:
On the error, on possibilty is that the allocated stack for vtables is exhausted. If it somehow believes a field id is higher than the last known, there might not be space on the stack, and it would crash, or fail on assert.

It could also be that a frame is not exited properly so the next frame exit operation asserts on the wrong frame type.

If you can give the assert line and comment from that line in builder, I might be able to give more hints.

skhoroshavin:
I've got working test case. In fact, I've got lots of working test cases, and some of them started to eat away all available memory so that I had to reboot my laptop :) I'll double check everything and commit soon.

mikkelfj:
never do anything half :)

PR #28, opened 2016-12-16
More tests

skhoroshavin:
Actually not a lot of new tests, but fixed typos in old ones that allowed them to pass. Now these tests pass since quick fix is still in effect, but if you comment [this](https://github.com/dvidelabs/flatcc/blob/master/src/compiler/codegen_c_json_parser.c#L565) line, then test will enter endless loop starting [there](https://github.com/dvidelabs/flatcc/commit/4d82573ea2f8a927b21c85e8677074d2f033cd7a#diff-9049181caf7ad9479b155c9c554775e2R199).

### DIFF
--- a/test/json_test/test_json.c
+++ b/test/json_test/test_json.c
@@ -196,15 +196,23 @@ int error_case_tests()
                 flatcc_json_parser_error_unknown_symbol );
     TEST_ERROR( "{ name: \"Monster\", test_type: Monster, test: { name: \"Joker\", colour: \"Red\" } } }",
                 flatcc_json_parser_error_unknown_symbol );
-    TEST_ERROR( "{ name: \"Monster\", test_array_of_tables: [ { nickname: \"Joker\", color: \"Red\" } ] }",
+    TEST_ERROR( "{ name: \"Monster\", testarrayoftables: [ { nickname: \"Joker\", color: \"Red\" } ] }",
                 flatcc_json_parser_error_unknown_symbol );
-    TEST_ERROR( "{ name: \"Monster\", test_array_of_tables: [ { name: \"Joker\", colour: \"Red\" } ] }",
+    TEST_ERROR( "{ name: \"Monster\", testarrayoftables: [ { name: \"Joker\", colour: \"Red\" } ] }",
                 flatcc_json_parser_error_unknown_symbol );
-    TEST_ERROR( "{ name: \"Monster\", test_array_of_tables: [ { name: \"Joker\", color: \"Red\", "
-                "test_type: Monster, test: { nickname: \"Harley\", color: \"Blue\" } } ] }",
+    TEST_ERROR( "{ name: \"Monster\", testarrayoftables: ["
+                "{ name: \"Joker\", color: \"Red\", test_type: Monster, test: { nickname: \"Harley\", color: \"Blue\" } } ] }",
                 flatcc_json_parser_error_unknown_symbol );
-    TEST_ERROR( "{ name: \"Monster\", test_array_of_tables: [ { name: \"Joker\", color: \"Red\", "
-                "test_type: Monster, test: { name: \"Harley\", colour: \"Blue\" } } ] }",
+    TEST_ERROR( "{ name: \"Monster\", testarrayoftables: ["
+                "{ name: \"Joker\", color: \"Red\", test_type: Monster, test: { name: \"Harley\", colour: \"Blue\" } } ] }",
+                flatcc_json_parser_error_unknown_symbol );
+    TEST_ERROR( "{ name: \"Monster\", testarrayoftables: ["
+                "{ name: \"Joker\", test_type: Monster, test: { nickname: \"Harley\" } },"
+                "{ name: \"Bonnie\", test_type: Monster, test: { name: \"Clyde\" } } ] }",
+                flatcc_json_parser_error_unknown_symbol );
+    TEST_ERROR( "{ name: \"Monster\", testarrayoftables: ["
+                "{ name: \"Joker\", test_type: Monster, test: { name: \"Harley\" } },"
+                "{ name: \"Bonnie\", test_type: Monster, test: { nickname: \"Clyde\" } } ] }",
                 flatcc_json_parser_error_unknown_symbol );
 
     return ret;


skhoroshavin:
Sorry for misguiding, endless loop starts here: https://github.com/skhoroshavin/flatcc/blob/master/test/json_test/test_json.c#L199

mikkelfj:
My guess is that the problem is cause by incorrect state management. is_table might not be correct in the given context. I seem to recall having had some problems managing this.

The error likely doesn't bail out immediately because it wants to clean something up, and if the state is incorrect chaos follows.

mikkelfj:
If debugging, add some breakpoints to detect the exact location where the error is being set, and see how the state looks at that point. Errors probably do not need cleanup, but likely can follow same code path till exit, if the state is correct. But can't dig deeper into this right now.

skhoroshavin:
Seems like error wasn't checked after loading table in an array, and everything tried to continue as if nothing happened, wrecking chaos in process. So I guess my fix is at least partially correct, since it adds error check exactly after returning from table parser in array parsing loop. As long as tests are passing and I don't have any problems with original case in my project I think this bug can be considered fixed. Would be great if you merged this PR as it adds working tests for this error, so that it doesn't come back.

skhoroshavin:
Btw, it seems like a good idea to create a harness for [coverage guided fuzzing](http://llvm.org/docs/LibFuzzer.html) to test for unexpected behavior in both json parser and verifier/reader. While I'm not up for this task now it could be added to issues to be addressed later.

mikkelfj:
On fuzz tester, I'm not likely to invest time in such a setup and issue would then create false expectations. I don't believe there at that many more bugs left to weed out, but I definitely appreaciate you work on improving tests and finding bugs.

On error fix, I'm not confident in anything complex that appears to be working. I need to understand what is happening, then I'll consider it fixed. Even if fixed correctly, we might have similar problems structs, enums etc.

mikkelfj:
BTW: one test case that would be good to have is parsing of recursively nested unions with different combinations of order and absence of type fields and object fields, and error line number handling.

mikkelfj:
I believe I have found the problem.
I'll give a post mortem to be better understand possible future errors.

In test_json/generated/monster_test_json_parser.h:1419
```
buf = MyGame_Example_Monster_parse_json_table(ctx, buf, end);
ref = flatcc_builder_end_table(ctx->ctx);
```
The invariant of the parser is to always return end if there is an error. This avoids checking for errors constantly because the end needs to be checked anyway. Often there is no need to return on error or `goto failed` because overall generated object parser catches the error when parse step completes.

The original error happens due to the test case setting up an error by inducing an unknown field with the runtime flag requiring unknown fields to be an error. The error is detected through many calls, but eventually the above two lines execute.

The first line (1419) sets buf to end because there is an error. The second call operates on invalid state because it does not check for the error. This is probably because the end table call does not consume input and therefore doesn't naturally check for `end`. Therefore, this is one of the cases where `goto failed` should be applied.

The infinite loop appears to be triggered by the buf value being reset to the buffer start repeatedly, for reasons I haven't fully investigated, but entering the next iteration of the object parser loop results in line 409      `buf = flatcc_json_parser_space(ctx, buf + 1, end);` which is bad because buf + 1 may be invalid, which is is. However, it is fair to expect here is no error when entering this part of the code because more is always reset on error. Somehow `more` has been set again, likely because we didn't bail at line 1420. This somehow cause the updated buffer to reset to the buffer start and failing the subsequent test line 410 `if (buf == end) {`

So why the infinite loop?
It appears that `end` is placed at zero termination before the buffer start, which is wrong. This fools the `flatcc_json_parser_space_ext` function which assumes end >= buf for its error handling. Thus it assumes no error and returns the buf + 1 as the new buffer, and the process repeats infinitely.

So the original error at line 1919 may not be the real error after all since it is fair to close a table after a failed parse assuming the parse loop will terminate.

Continue investigation.

mikkelfj:
It is worth noting that the quick fix in #27 tests for error after line 1419, but of course fixes the logic in the code generator, not in the generated json test parser.

mikkelfj:
After further analysis:
The vector loop in the parser has two invariants related to errors: ref == 0 or buf == end.
The parser avoids an error to travel far, trading overhead in case of error with overhead in typical case.
Therefore, nearly all functions are robust to errors and all a logic segment needs to do, is to maintain this invariant.

The array loop has a general bug. It does not observe the buf == end and keeps looping. The error handler logic solves this by setting more = 0, but this has no effect, possibly due to changes in the loop logic at some point or another.

Note that more is reused in two loops, but more is always updated towards the end of the loop making such reuse possible, and also causes the error bailout to fail.

Ensuring more is reset in the following, when buf == end` should handle all cases that the quick fix solution does not handle:
`buf = flatcc_json_parser_array_end(ctx, buf, end, &more);`

The quickfix adds an apparently unnecessary test for error at the mentioned line 1419 (generated). However, in rare runtime error cases, the builder may not have entered the table before the error was discovered. It is almost not worth while handling, but should be added. The ideomatic way to handle the quick fix is `if (buf == end) goto failed` instead of `if (ctx->error) return buf`. This is faster because it avoids a pointer indirection, and although not needed here, failed ensures a runtime error is set if no other error has been raised, and we do not care about performance when failing. Later code changes might rely on this.

As it turns out, the code already does this correctly, except the quick fix very very special case, but it has nothing to do with the actual error, is just short-circuits the problem in one special case, failing to deal with all other cases.

As we investigate further, similar to array_end, object_end also sets more = 0 when buf == end. Thus, our error continue to propagate as it should.

Until finally we reach `flatcc_json_parser_finalize_unions` at the very end of a table parser that contains unions.

This function fails to to test for buf == end which is the true reason for the error.
But, because if the very rare special case mentioned above, we still need a variation of the quick fix. It is almost not worthwhile, but we can expect the compiler to shortcircuit the test logic and remove the overhead, especially if we use the ideomatic `buf == end` test.

mikkelfj:
Quick summary:

The real problem was `flatcc_json_parser_finalize_unions` which failed to propagate the error when `buf == end`. The quick fix only solved a special case in an unorthodox way, but accidentally happened to fix another extremely rare error condition so it will kept in a modified version.

#30 is the issue tracking this bug

mikkelfj:
This wasn't quite right (as it was tested with quick fix enabled), the end_array did in fact fail to check for end like object end does. Finalize unions still needs the check. When fixing this, the function correctly bails out and passes the test case at line 199 in json_test.c, also if the quick fix is disabled.

However, the test then fails two test cases down with an incorrect frame assertion:
`-> 1407     check(frame(type) == flatcc_builder_offset_vector, "expected offset vector");`
```
-> 203      TEST_ERROR( "{ name: \"Monster\", testarrayoftables: ["                                                    │
   204                  "{ name: \"Joker\", color: \"Red\", test_type: Monster, test: { nickname: \"Harley\", color: \"│
Blue\" } } ] }",
```

mikkelfj:
After more debugging:
unions handle tables a bit like the quick fix: they only call table_end `if (buf != end)`. In test case at line 203, we have a table that contains a union which is another table. With an error in the inner, union, table, the parent table (without quick fix) would end the table of the union instead of its own, which is ok, but the vector wrapping this higher up table will now see a table during cleanup. There are several ways to handle this, but early out is one of them which the quick fix supports and which is consistent with union handling. Because offset vectors only hold strings and simple tables, this error case is covered by the quick fix. However, the array_end must still be patched to handle parse errors of other types such as strings, structs, and scalars.

In conclusion, while the quick fix did solve the problem it did not solve all cases, but it did solve two other cases: unions failing, and unlikely failing to create table start in builder (alloc failure). We do not have test cases covering the new cases I discovered,  but I'll commit a patch to master with these changes.

skhoroshavin:
Good to hear you looked into this and fixed more erroneous cases. It would take quite a long time for me to do this research.

PR #29, opened 2016-12-17
Allow to generate scan for all fields as a default option

skhoroshavin:
We discussed possibility of adding scan functions for all fields as a default option and I tried to implement it.  I also tried in my project to remove keys from fields, regenerate headers with new flatcc and measure compilation time on my laptop. After several clean runs of "time make all" it became apparent that there is no noticeable effect on compilation time (7.5 sec before and after on average) nor binaries size (514K executable compiled in debug mode). Reader header size became even smaller (21K before, 20K after).

### DIFF
--- a/test/monster_test/monster_test.c
+++ b/test/monster_test/monster_test.c
@@ -1109,7 +1109,7 @@ static size_t count_monsters(ns(Monster_vec_t) monsters, const char *name)
 
     for (i = ns(Monster_vec_scan)(monsters, name);
          i != nsc(not_found);
-         i = ns(Monster_vec_scan_from)(monsters, i + 1, name)) {
+         i = ns(Monster_vec_scan_ex)(monsters, i + 1, nsc(end), name)) {
         ++count;
     }
 
@@ -1173,7 +1173,7 @@ int test_scan_by(flatcc_builder_t *B)
         printf("scan_by did not find the Joker with n\n");
         goto done;
     }
-    if (nsc(not_found) != ns(Monster_vec_scan_from(monsters, 2, "Joker"))) {
+    if (nsc(not_found) != ns(Monster_vec_scan_ex(monsters, 2, nsc(end), "Joker"))) {
         printf("scan_from found Joker past first occurence\n");
         goto done;
     }
@@ -1189,32 +1189,32 @@ int test_scan_by(flatcc_builder_t *B)
         printf("Gulliver not found\n");
         goto done;
     }
-    if (2 != ns(Monster_vec_scan_from_by_name(monsters, 2, "Gulliver"))) {
+    if (2 != ns(Monster_vec_scan_ex_by_name(monsters, 2, nsc(end), "Gulliver"))) {
         printf("Gulliver not found starting from Gulliver\n");
         goto done;
     }
-    if (4 != ns(Monster_vec_scan_from_by_name(monsters, 3, "Gulliver"))) {
+    if (4 != ns(Monster_vec_scan_ex_by_name(monsters, 3, nsc(end), "Gulliver"))) {
         printf("Another Gulliver not found\n");
         goto done;
     }
 
-    if (nsc(not_found) != ns(Monster_vec_scan_range(monsters, 1, 3, "Jingle"))) {
+    if (nsc(not_found) != ns(Monster_vec_scan_ex(monsters, 1, 3, "Jingle"))) {
         printf("not found in subrange not working\n");
         goto done;
     }
-    if (nsc(not_found) != ns(Monster_vec_scan_range(monsters, 1, 3, "TwoFace"))) {
+    if (nsc(not_found) != ns(Monster_vec_scan_ex(monsters, 1, 3, "TwoFace"))) {
         printf("subrange doesn't limit low bound\n");
         goto done;
     }
-    if (1 != ns(Monster_vec_scan_range(monsters, 1, 3, "Joker"))) {
+    if (1 != ns(Monster_vec_scan_ex(monsters, 1, 3, "Joker"))) {
         printf("scan in subrange did not find Joker\n");
         goto done;
     }
-    if (2 != ns(Monster_vec_scan_range_by_name(monsters, 1, 3, "Gulliver"))) {
+    if (2 != ns(Monster_vec_scan_ex_by_name(monsters, 1, 3, "Gulliver"))) {
         printf("scan in subrange did not find Gulliver\n");
         goto done;
     }
-    if (nsc(not_found) != ns(Monster_vec_scan_range_by_name(monsters, 1, 3, "Alice"))) {
+    if (nsc(not_found) != ns(Monster_vec_scan_ex_by_name(monsters, 1, 3, "Alice"))) {
         printf("subrange doesn't limit upper bound\n");
         goto done;
     }
@@ -1245,7 +1245,7 @@ int test_scan_by(flatcc_builder_t *B)
         printf("scan not working on middle item of inventory\n");
         goto done;
     }
-    if (nsc(not_found) != (pos = nsc(uint8_vec_scan_from(inv, 3, 1)))) {
+    if (nsc(not_found) != (pos = nsc(uint8_vec_scan_ex(inv, 3, nsc(end), 1)))) {
         printf("scan_from(item+1) not working on middle item of inventory\n");
         goto done;
     }
@@ -1261,11 +1261,11 @@ int test_scan_by(flatcc_builder_t *B)
         printf("scan not working for repeating item of inventory\n");
         goto done;
     }
-    if (3 != (pos = nsc(uint8_vec_scan_from(inv, 3, 3)))) {
+    if (3 != (pos = nsc(uint8_vec_scan_ex(inv, 3, nsc(end), 3)))) {
         printf("scan_from(item) not working for repeating item of inventory\n");
         goto done;
     }
-    if (5 != (pos = nsc(uint8_vec_scan_from(inv, 4, 3)))) {
+    if (5 != (pos = nsc(uint8_vec_scan_ex(inv, 4, nsc(end), 3)))) {
         printf("scan_from(item+1) not working for repeating item of inventory\n");
         goto done;
     }
--- a/test/monster_test/monster_test.c
+++ b/test/monster_test/monster_test.c
@@ -1169,6 +1169,10 @@ int test_scan_by(flatcc_builder_t *B)
         printf("scan_by did not find the Joker\n");
         goto done;
     }
+    if (1 != ns(Monster_vec_rscan(monsters, "Joker"))) {
+        printf("rscan_by did not find the Joker\n");
+        goto done;
+    }
     if (1 != ns(Monster_vec_scan_n(monsters, "Joker3", 5))) {
         printf("scan_by did not find the Joker with n\n");
         goto done;
@@ -1186,6 +1190,10 @@ int test_scan_by(flatcc_builder_t *B)
         goto done;
     }
     if (2 != ns(Monster_vec_scan_by_name(monsters, "Gulliver"))) {
+        printf("Gulliver not found\n");
+        goto done;
+    }
+    if (4 != ns(Monster_vec_rscan_by_name(monsters, "Gulliver"))) {
         printf("Gulliver not found\n");
         goto done;
     }
--- a/test/monster_test/monster_test.c
+++ b/test/monster_test/monster_test.c
@@ -5,6 +5,7 @@
 
 #include "flatcc/support/hexdump.h"
 #include "flatcc/support/elapsed.h"
+#include "../../config/config.h"
 
 /*
  * Convenience macro to deal with long namespace names,
@@ -1116,12 +1117,13 @@ static size_t count_monsters(ns(Monster_vec_t) monsters, const char *name)
     return count;
 }
 
-int test_scan_by(flatcc_builder_t *B)
+int test_scan(flatcc_builder_t *B)
 {
     size_t pos;
     ns(Monster_table_t) mon;
     ns(Monster_vec_t) monsters;
     nsc(uint8_vec_t) inv;
+    nsc(string_vec_t) strings;
     void *buffer;
     size_t size;
     uint8_t invdata[] = { 6, 7, 1, 3, 4, 3, 2 };
@@ -1131,6 +1133,9 @@ int test_scan_by(flatcc_builder_t *B)
     ns(Monster_start_as_root(B));
     ns(Monster_name_create_str(B, "MyMonster"));
     ns(Monster_inventory_create(B, invdata, c_vec_len(invdata)));
+
+    ns(Monster_testarrayofstring_start(B));
+    ns(Monster_testarrayofstring_end(B));
 
     ns(Monster_testarrayoftables_start(B));
 
@@ -1164,6 +1169,8 @@ int test_scan_by(flatcc_builder_t *B)
     assert(monsters);
     inv = ns(Monster_inventory(mon));
     assert(inv);
+    strings = ns(Monster_testarrayofstring(mon));
+    assert(strings);
 
     if (1 != ns(Monster_vec_scan(monsters, "Joker"))) {
         printf("scan_by did not find the Joker\n");
@@ -1174,6 +1181,10 @@ int test_scan_by(flatcc_builder_t *B)
         goto done;
     }
     if (1 != ns(Monster_vec_scan_n(monsters, "Joker3", 5))) {
+        printf("scan_by did not find the Joker with n\n");
+        goto done;
+    }
+    if (1 != ns(Monster_vec_rscan_n(monsters, "Joker3", 5))) {
         printf("scan_by did not find the Joker with n\n");
         goto done;
     }
@@ -1197,7 +1208,19 @@ int test_scan_by(flatcc_builder_t *B)
         printf("Gulliver not found\n");
         goto done;
     }
+    if (4 != ns(Monster_vec_rscan_n_by_name(monsters, "Gulliver42", 8))) {
+        printf("Gulliver not found with n\n");
+        goto done;
+    }
+    if (2 != ns(Monster_vec_rscan_ex_n_by_name(monsters, 1, 3, "Gulliver42", 8))) {
+        printf("Gulliver not found with n\n");
+        goto done;
+    }
     if (2 != ns(Monster_vec_scan_ex_by_name(monsters, 2, nsc(end), "Gulliver"))) {
+        printf("Gulliver not found starting from Gulliver\n");
+        goto done;
+    }
+    if (2 != ns(Monster_vec_scan_ex_n_by_name(monsters, 2, nsc(end), "Gulliver42", 8))) {
         printf("Gulliver not found starting from Gulliver\n");
         goto done;
     }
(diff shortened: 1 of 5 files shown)


mikkelfj:
This is also what I would expect from a modern compiler. What is less clear is what happens with some embedded systems compiler. That said, I already decided to rely heavily on unused inlines being essentially free and there is now a flag control it.

I would probably have negated the flag test and moved the scan into an existing macro implementing scalar and string vector fields when the flag is set but probably not worth it.

mikkelfj:
BTW: it would be good with a monster_test that scans a vector that is not key, when flag is on.

skhoroshavin:
> This is also what I would expect from a modern compiler. What is less clear is what happens with some embedded systems compiler. That said, I already decided to rely heavily on unused inlines being essentially free and there is now a flag control it.

Considering embedded systems. While I admit that there are some popular architectures not supported by gcc (namely 8051), they are declining in popularity. ARM Cortex (supported by gcc pretty well) is becoming increasingly more popular and diverse - you can buy low-end Cortex-M0 MCU for a price comparable to 8051 MCU, and many major players are superseding their 8051-based products with ARM ones. One example is quite popular TI CC2541 8051 SoC with embedded bluetooth capability, which now have similarly priced successor CC2640 based on ARM Cortex. Another popular architecture is AVR, and it's also supported by gcc very well. Considering other compilers - I had experience with sdcc, it's quite dumb, but even this compiler eliminates unused inline functions as far as I remember. And another quite popular compiler for embedded is IAR, and from what I heard about it it's on par with gcc. So, I think your assumption of unused inlines being free is right for almost any system.

> I would probably have negated the flag test and moved the scan into an existing macro implementing scalar and string vector fields when the flag is set but probably not worth it.

I was thinking about this, but decided not to implement it now so that command-line option could be added if needed. But if you wish I can do this move.

> BTW: it would be good with a monster_test that scans a vector for a non-key field, when flag is on.

Okay.

mikkelfj:
> Considering embedded systems.
The point is to support just about everything - people do use it for esoteric systems.

> I was thinking about this, but decided not to implement it now so that command-line option could be added if needed. But if you wish I can do this move.

If it reduces noise level in non-common files I think it is a worthwhile change. I see the point of cmd args, but there are several flagged features that all might start to bloat if done that way.

Note that there are also runtime flags, but probably not relevant here:
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_rtconfig.h

mikkelfj:
After looking at the generated output, I would definitely say it should be placed in a macro. Please do this and add the test, which also serves as an example when other documentation is absent.

If we add more detailed cmd args, as the system is prepared for, we need to rethink many things other than the scan feature, so best to just follow the current approach.

BTW: I was wondering if there is any reason other than sloppyness that the is_present method is generated in the schema specific code, since you are deeper into the source than I am atm.

Here is a sample of the generated output in it's current form in this PR

```
static inline flatbuffers_bool_t MyGame_Example_Monster_testbool(MyGame_Example_Monster_table_t t)
__flatbuffers_scalar_field(flatbuffers_bool, 15, 1, t)
__flatbuffers_define_scan_by_scalar_field(MyGame_Example_Monster, testbool, flatbuffers_bool_t)
static inline int MyGame_Example_Monster_testbool_is_present(MyGame_Example_Monster_table_t t)
__flatbuffers_field_present(15, t)
```

skhoroshavin:
> After looking at the generated output, I would definitely say it should be placed in a macro. Please do this and add the test, which also serves as an example when other documentation is absent.

Okay.

> I was wondering if there is any reason other than sloppyness that the is_present method is generated in the schema specific code, since you are deeper into the source than I am atm.

I think the reason is that it's only used for table fields, not structs where all fields are always present.

On the naming of scan functions. I thought about scan_n_from_by_name and scan_from_n_by_name, and both are not ideal. But what about scan_n_by_name_from? It's both readable and plays nice with code completion.

mikkelfj:
> On the naming of scan functions. I thought about scan_n_from_by_name and scan_from_n_by_name, and both are not ideal. But what about scan_n_by_name_from? It's both readable and plays nice with code completion.

Not directly opposed, but wouldn't it go against current conventions for limited gain? We can change the entire API deprecating names if it is really beneficial and I'd rather do that or live with slightly ugly names over inconsistency.

skhoroshavin:
Not sure it's really inconsistent. When comparing to find - currently we only have find_by_name, find_n_by_name, find and find_n. There are no other cases where we have ending "by_name" except newly added scan.

mikkelfj:
I see, but it still places argument context on both sides of by_name, and I'm wondering about the precedence for other operations. Intuively it should be placed before the 'n', and the _by_name is a suffix applied to vector operations. So suddenly you insert a _by_name in the middle of a vector operation. `uint8_vec_scan_from_n` vs  `Monster_id_scan_n_by_name_from` (from memory)

skhoroshavin:
On the other hand there are many cases with field accessors ending with field name, but there is also name_is_present function. Also from the point of code completion it could be quite good to type scan, see numerous by_name variants, continue with by_name, and either stop or see options to do scan from or scan range.

Also, what about dropping _n variants? Or renaming them into just scan, and allowing to pass -1 as string length meaning that it should be calculated automatically? Or moving _n after by_name - scan_by_name_n, making it more clear that n is for name data format, not anything else?

mikkelfj:
The -1 arguments is convenient from an api perspective, but it is really annoying from a user perspective. It is not many days ago you proposed `range` should not exist because `from` was the typical use case. We ended up with both as both are useful.

`is_present` couldn't really be any different as it is actually an operation on the field, and regardless we always have to weigh the alternatives as any choice has downsides.

The reason the vector operations are before the name (_by_name), is probably because a) the are a suffix to the equivalent vector operations, b) they are vector operations, and not operations on the field itself.

mikkelfj:
Another way to see this, is to consider the original default find operation: `Monster_find`. It finds monsters. `Monster_scan_from_by_name` scans monsters from a given position by comparing a given field against a key. It is not the key field that is taken from a given range or position.

BTW: I'm not saying the naming scheme is perfect or the only possible, but I thought long and hard about the various options, and would have appreciated a discussion like this. So I try explain the logic behind the naming while being open to suggestions.

mikkelfj:
On moving n to the end, I would almost buy that argument (I think I suggested it myself at one point not longer ago), but it breaks with the standalone vector operations. For one thing, it is convenient to search for operations in text editors, and if the name suddenly splits up, this is no longer possible. And it becomes confusing for potential code generators  - e.g. language interfaces relying on generated C code.

skhoroshavin:
> The -1 arguments is convenient from an api perspective, but it is really annoying from a user perspective. It is not many days ago you proposed range should not exist because from was the typical use case. We ended up with both as both are useful.

Agreed :) Well, dropping _n was just one of many variants.

From readability POV scan_from_by_name looks a bit awkward, but scan_range_by_name looks perfect. And while scan_by_name_from reads better, scan_by_name_range is not quite. And there is consistency issue, so on average scan_from_by_name/scan_range_by_name looks more preferable (maybe finding some noun instead of from would make this perfect?)

> On moving n to the end, I would almost buy that argument (I think I suggested it myself at one point not longer ago), but it breaks with the standalone vector operations. For one thing, it is convenient to search for operations in text editors, and if the name suddenly splits up, this is no longer possible. And it becomes confusing for potential code generators - e.g. language interfaces relying on generated C code.

Good point.

mikkelfj:
>  (maybe finding some noun instead of from would make this perfect?)

I was thinking the same while reading through your comment.

Maybe we should be developing on a feature branch, I don't really like to break an api multiple times even if not yet released, but here we need to get this right. Though, I think we will be closing this soon.

BTW: I keep thinking (and have done so in the past), that find and sort could also be an other fields, but I keep backing away because it seems kind of bloated, and because sort isn't really a normal operation on a finalized buffer - for one thing it is potentially unsafe if you didn't create it in a DMZ.

skhoroshavin:
> I was thinking the same while reading through your comment.

One more (maybe stupid) idea - drop from and revive next. While from somewhat more flexible than next, I suspect that there would be not that many situations where one would really need from and not range.

> Maybe we should be developing on a feature branch, I don't really like to break an api multiple times even if not yet released, but here we need to get this right. Though, I think we will be closing this soon.

Not sure about branching, I hope this could be closed before holidays.

mikkelfj:
We definitely shouldn't go back to `next` - I think the examples showed that. But I am wondering if `scan_at` is preferable after all? Or `scan_right`, `scan_left`. I think it will end with `scan_from`, but keep'em coming.

skhoroshavin:
Looked through previous discussion and got one more idea - scan_by_name_from/scan_by_name_in? instead of range.

mikkelfj:
Again, its after by_name. We should, btw, reserve a reverse search variant, such as rscan_range or scan_rrange. With this in mind, at sounds better. rscan_at(monsters, Monster_vec_len(monsters), "Joker").

skhoroshavin:
> Again, its after by_name. We should, btw, reserve a reverse search variant, such as rscan_range or scan_rrange. With this in mind, at sounds better. rscan_at(monsters, Monster_vec_len(monsters), "Joker").

Okay, searching for nouns. Btw, I don't see at being any better in this situation - in fact rscan_from is still more readable - it does imply that we scan starting from some position, but makes no assumption on direction.

skhoroshavin:
Looked through several online dictionaries to find appropriate synonyms - no luck. Seems like we'd better stick with current scan_from_by_name/scan_n_from_by_name.

mikkelfj:
Yes. But didn't we agree to move _n after from/range? I thought you'd commit such a change?

skhoroshavin:
I remember we were hesitating whether scan_n_from or scan_from_n would be better. For scan_from_n is easier to find, but it's name is quite misleading.

skhoroshavin:
And ironically, scan_from_n_by_name does look good for describing scan starting from some positions :) Or scan_from_pos_by_name.

mikkelfj:
`scan_from` could also be `scan_partial`, `scan_part`, `scan_offset`, `scan_reminder`, `scan_more`, `scan_tail`.

`scan_range` could also be `scan_slice`, `scan_segment`, `scan_limit`, `scan_sub`.

mikkelfj:
and of course, `scan_less` :) - personally I don't think `scan_part` is all that bad if we are looking for nouns.

skhoroshavin:
scan_tail looks best so far :) and rscan_tail is also not that bad.

mikkelfj:
Of course mathematically, it would be `scan_half_bounded_interval` and `scan_interval`, or `scan_half_segment`, `scan_segment`.

But I also like tail.

https://en.wikipedia.org/wiki/Interval_(mathematics)

skhoroshavin:
So, it seems like finally we have an agreement: scan_n_by_name/scan_tail_n_by_name/scan_range_n_by_name

mikkelfj:
I was about to write the same.

mikkelfj:
A completely other option, invented by microsoft, which I use on occasion, is an _ex function that takes everything you can throw at it. It could be `scan[_n][_by_<name>](vec, key)` and `scan[_n]_ex[_by_<name>](vec, first, last, inc, key)` where inc is signed.

mikkelfj:
Or `scan[r][_n]_ex[_by_<name>](vec[, first, last], key[, key_len])` which is effectively the same as dropping tail. Not saying we should, but it is an option.

skhoroshavin:
Interesting idea, but what about consistency with find? Also, we drop tail variant, which isn't desirable due to it's use in loops.

mikkelfj:
It is not inconsistent with find, because we can add find_ex trivially. For loops, we would always use the ex variant and just add `flatbuffers_end` as argument for last.

mikkelfj:
btw, I mean `[r]scan`, not `scan[r]`:

`[r]scan[_n]_ex[_by_<name>](vec[, first, last], key[, key_len])`

mikkelfj:
```
/* flatbuffers_end == `(size_t)-1` == `flatbuffers_not_found` */
size_t a, b = flatbuffers_end;
const char *name = "Joker";

a = Monster_scan(monsters, name);
while (a < b) {
    a = Monster_scan_ex(monsters, a + 1, b, name);
}
```

skhoroshavin:
```
for (i = monster_vec_scan_by_name(m, "Joker");
      i != nsc(not_found);
      i = monster_vec_scan_ex_by_name(m, i + 1, nsc(end), "Joker"))
{
...
}
```

skhoroshavin:
Looks ok. So, going ahead to implement it?

mikkelfj:
yeah, it is a lot simpler that guessing names all the time.

Note that nsc makes it possible to have alternative flatbuffers such as flatbuffers_wide_end for 64-bit, but for normal end-user doc `flatbuffers_end` would be used, possibly with a comment about the namespace could be `flatbuffers_<alternative>_` also. nsc is used when you need independence of this (and for historical reasons).

mikkelfj:
Why don't you throw in rscan now you at it, it's going to happen at some point anyway. It's useful for sorted scan.

mikkelfj:
Ouch, how do you delimit rscan at index 0?

mikkelfj:
Is it always first < last for non-empty ranges, and rscan searches in last-1 thorugh first, and scan search first through last-1? What does the loop look like?

skhoroshavin:
```
for (i = monster_vec_rscan_by_name(m, "Joker");
      i != nsc(not_found);
      i = monster_vec_rscan_ex_by_name(m, 0, i, "Joker"))
{
...
}
```
note absence of -1, could be misleading

mikkelfj:
How misleading?

mikkelfj:
I think scan_[n]_ex is better, even though I would argue opposite if the name was range. I would never put a suffix on a normal _ex function, but _by_name is an exception. _n I'm less sure about?

skhoroshavin:
Copy-paste forward loop, replace scan with rscan, and it could be quite easy to forget that i+1 should be changed to i, not i-1. But maybe I'm just overworrying.

skhoroshavin:
I think searchwise scan_ex_n is still more preferable. And we already have _by_name exception.

mikkelfj:
> Copy-paste forward loop, replace scan with rscan, and it could be quite easy to forget that i+1 should be changed to i, not i-1. But maybe I'm just overworrying.

Definitely a source of bugs, but such is life.

I still doesn't sit right with me on the ex thing, give it some thought overnight.

skhoroshavin:
> I still doesn't sit right with me on the ex thing, give it some thought overnight.

Okay.

I started to implement rscan, and I have a problem.
```
#define __flatbuffers_rscan_by_field(b, e, A, V, E, L, K, Kn, T, D)\
{ T v; size_t i;\
  for (i = e-1; i >= b; --si) {\
    v = A(E(V, i));\
    if (D(v, (K), (Kn)) == 0) {\
       return i;\
    }\
  }\
  return flatbuffers_not_found;\
}
```

```
        "static inline size_t N ## _vec_rscan_by_ ## NK(N ## _vec_t vec, TK key)\\\n"
        "__%srscan_by_scalar_field(0, N ## _vec_len(vec), N ## _ ## NK, vec, N ## _vec_at, N ## _vec_len, key, TK)\\\n"
```

And now I'm getting warning (which is error actually) that `i >= b` is always true.

mikkelfj:
```
/* flatbuffers_end == `(size_t)-1` == `flatbuffers_not_found` */
size_t a = 0, b = flatbuffers_end, n, count = 0;
const char *name = "Joker";
n = strlen(name);

#if 0

do {
    b = Monster_vec_rscan_n_ex(monsters, a, b, name, n);
} while (b != nsc(not_found) && ++count);

#else

do {
    a = Monster_vec_rscan_ex_n(monsters, a, b, name, n);
} while (b != nsc(not_found) && ++count);

#endif
```

mikkelfj:
> And now I'm getting warning (which is error actually) that i >= b is always true.

You are decrementing si, but testing i, and you do not guard for `b == 0`.

skhoroshavin:
Well, si was just a typo, and it didn't fix anything. As for guarding for b==0 - is this really needed?

mikkelfj:
Yeah, I assumed it was a type, but you still don't guard for b == 0, try this, for loops are not always best.

```
#define __flatbuffers_rscan_by_field(b, e, A, V, E, L, K, Kn, T, D)\
{ T v; size_t i = e;\
  while (i-- > b) {
    v = A(E(V, i));\
    if (D(v, (K), (Kn)) == 0) {\
       return i;\
    }\
  }
  return flatbuffers_not_found;\
}
```

mikkelfj:
More detailed, if b is 0

`for (i = e-1; i >= b; --i) {\`

you get i = SIZE_MAX if i is 0, otherwise you loop until i is 0, then i becomes size max, so you never terminate the loop. If b is non-zero, it will work. My loop variant test the precondition is larger which does not have that problem. This is why Go (lang) insists on using signed integers. But compiler optimizations are much worse due to some absent guarantees on overflow.

mikkelfj:
signing off for tonight

skhoroshavin:
Thanks, seems like I need to get some sleep too.

mikkelfj:
To summarize, I'm waiting for test cases, move into field macro, and decision on _n placement.

Anything else outstanding before a release?
I've got @csstup standby for testing big endian which will also take some time.

test/monster_test/monster_test.c
```diff
@@ -5,6 +5,7 @@
 
 #include "flatcc/support/hexdump.h"
 #include "flatcc/support/elapsed.h"
+#include "../../config/config.h"
```
mikkelfj: See if #include "config.h" works, config should  be in path. There are a few places with ../../ but it is best avoided.

skhoroshavin: I already tried "config.h" but no luck. I could look into cmakelists for monster_test and try to add it to include path, so that ../.. is not needed - I don't like current solution either, just wasn't unsure if modifying build parameters here is acceptable, given other places with ../..

As for rsync - I'll add them in couple of minutes

skhoroshavin:
Sorry for being suddenly offline - had some problems to solve. Now back at flatcc. I've added some tests for non-key fields, not sure if they are good enough or not. At least they check that scan and scan_ex exist and somewhat work. If more thorough testing is needed I would consider rethinking monster test architecture to make it more generic.

As for _n placement. As far as I see we have 3 options:
1. scan_n_ex_by_name
2. scan_ex_n_by_name
3. scan_ex_by_name_n

Option 3 was already dropped since there is also default scan without by_name, and this looks somewhat inconsistent. On the other hand, n is the last parameter and it's reflected by this naming quite nicely.

Option 2 looked quite natural for me after deciding on scan_range_n, and it makes search for scan_ex easier compared to option 1. And option 1 was proposed because in microsoft libraries _ex usually comes last, but this is already broken by by_name part, so I'm not sure if this argument makes 1 really better than 2.

Personally, I would prefer either 2 or 3, but I won't be much disappointed if you still decide on 1.

skhoroshavin:
Updated tests with rscan

mikkelfj:
> Option 2 looked quite natural for me after deciding on scan_range_n, and it makes search for scan_ex easier compared to option 1.

It sits better with me today than yesterday, but could easily see myself seaching for scan_n to fix some string len args. Let us keep 2) for now. Also, while I think _ex is probably the best I wouldn't entirely rule out range yet. Tests will clarify.

On quality of tests, I made a few review comments. But in summary, the flagged tests cannot stand alone, they just need to make sure the code compiles with the given function names. I have been reading through your tests cases, now and before, and you seem to capture the cases I would expect, except the reverse scan should find last of 2 keys. Range conditions are also important, rscan with 0 boundary.

skhoroshavin:
Tried to move define_scan_by_xxx_field into xxx_field macro - seems like it needs more attention than previously thought. I had to shuffle around generation of some macros and typedefs, and scalar_field macro needs to be changed as well. Currently field accesor definition looks like
```
static inline MyGame_Example_Color_enum_t MyGame_Example_TestSimpleTableWithEnum_color(MyGame_Example_TestSimpleTableWithEnum_table_t t)
__flatbuffers_scalar_field(MyGame_Example_Color, 0, 2, t)
```
with function signature generated explicitly and __flatbuffers_scalar_field is macro containing only function body.

Now __flatbuffers_scalar_field macro should also contain this:
```
__flatbuffers_define_scan_by_scalar_field(MyGame_Example_TestSimpleTableWithEnum, color, MyGame_Example_Color_enum_t)
```
and it's quite obvious that it just doesn't contain enough parameters to pass to define_scan. Also it becomes somewhat inconsistent containing function body _and_ another macro containing another function signature and body. I could add more parameters and move generation of accessor function signature into __flatbuffers_scalar_field macro, but I'm not sure if it's actually good idea.

mikkelfj:
Quick reply: it may need some reorganisation and more macro parameters. This has been done elsewhere, but I think you should be able to find similar approaches elsewhere - but it needs careful attention and thought, and shouldn't be rushed.

skhoroshavin:
> On quality of tests, I made a few review comments. But in summary, the flagged tests cannot stand alone, they just need to make sure the code compiles with the given function names. I have been reading through your tests cases, now and before, and you seem to capture the cases I would expect, except the reverse scan should find last of 2 keys. Range conditions are also important, rscan with 0 boundary.

I've reread both review comments and this reply, and I still don't get if I should do anything else with tests now :(

mikkelfj:
OK: we need coverage tests for scan and rscan for a known keyed field such as monster name. This test should cover all function name variants, common pitfalls such as dual keys, and some boundary conditions to show they work, e.g. search [0, 0), [len, end) should always be fail.

We already have most of this for scan, but not for rscan.

Test flagged with ALLOW_SCAN_ON_ALL_FIELDS (or something) are not reliable because they only take effect if the feature is activated. But, there need to be a smoke test to ensure at least some of the scan names are present on non-key fields when the flag is activated. (Testing the negative is not worthwhile since that requires a compiler error).

mikkelfj:
The above was a general requirement - tests do cover some, but not all of those requirements. E.g. rscan does not have a feature test outside the all fields feature as I recall.

mikkelfj:
I can have a look at the macro impl., but not tonight.

skhoroshavin:
Reduced tests for non-key fields, added more test cases.

mikkelfj:
Good, also need the _n variant, but just for smoke on the keyed field, or what you find useful.

skhoroshavin:
Added rscan_n variant on monster name (and tests already contain scan_n)

mikkelfj:
Must be exhausting ...,
but we need a test on all variants so they don't disappear from the api by mistake, making users unhappy, so nsc(string_vec_rscan_ex_n), Monster_vec_rscan_ex_n_by_name, scan_ex_n ditto where missing.

skhoroshavin:
Even more scan variants. Hope didn't miss anything.

mikkelfj:
Monster_vec_rscan_ex_n_by_name
I think the scan_ex_n_by_name is missing, but otherwise it looks ok,

skhoroshavin:
https://github.com/dvidelabs/flatcc/pull/29/commits/c17a12763d8e88d9551e6e1838a33f86fffaea3a#diff-c2ebe16717f17f52e0f3fa858f005325R1215

mikkelfj:
> c17a127#diff-c2ebe16717f17f52e0f3fa858f005325R1215

that is rscan_ex_n
scan_ex_n is missing, or am I mistaken?

test/monster_test/monster_test.c
```diff
@@ -1270,13 +1344,39 @@ int test_scan_by(flatcc_builder_t *B)
...
+    nsc(string_vec_scan(strings, "Hello"));
+    nsc(string_vec_scan_ex(strings, 0, nsc(end), "Hello"));
+    nsc(string_vec_scan_n(strings, "Hello", 4));
+    nsc(string_vec_scan_ex_n(strings, 0, nsc(end), "Hello", 4));
+    nsc(string_vec_rscan(strings, "Hello"));
+    nsc(string_vec_rscan_ex(strings, 0, nsc(end), "Hello"));
+    nsc(string_vec_rscan_n(strings, "Hello", 4));
+    nsc(string_vec_rscan_ex_n(strings, 0, nsc(end), "Hello", 4));
+
+#ifdef FLATCC_ALLOW_SCAN_FOR_ALL_FIELDS
```
mikkelfj: This should not be #ifdef

The configuration assigns 0 or 1 so it can be overridden by compile flag.
#if FLATCC_ALLOW_SCAN_FOR_ALL_FIELDS
...
#endif

mikkelfj:
See also review comment on #ifdef flag test

skhoroshavin:
Fixed ifdef, added one more test

mikkelfj:
Great, that should cover the scan feature.

As to moving into field macro. If there is only a single define on each field for scan, we could live with it for now, but I digged briefly into the macro system:

The builder has a more advanced field definition that should cover the missing parts. Whether it is worthwhile to follow this example for the reader, I'm not sure about.

But you can see the code in build/Debug/test/monster_test/generated/monster_test_builder.h

e.g.
```
__flatbuffers_build_scalar_field(20, flatbuffers_, MyGame_Example_Monster_testhashs32_fnv1a, flatbuffers_int32, int32_t, 4, 4, 0)
```

And the flatbuffers_common_reader/builder.h in same directory is easier to read than the code generator.

mikkelfj:
Field cleanup is better handled in seperate PR, so merging now.

mikkelfj:
Keeping this issue open until we figure out how/if to integrate scan ninto the field macroes in the reader interface.

PR #42, opened 2017-05-12
Produce a shared library

epouponstormshield:


mikkelfj:
While a appreciate the contribution, and recognize that this is probably needed, I am not convinced this is the right way to do it because the INSTALL target will automatically install on various systems and this might break convention.

There is a CMAKE flag to choose to either build a shared or a static library already:

`CMAKE ... -DBUILD_SHARED_LIBS=ON ...`

http://cgold.readthedocs.io/en/latest/tutorials/libraries/static-shared.html

I just want a good discussion on what is best to do, and I am no expert on this so I'll leave this open for a while for others to comment on. My concern is that some users might want to link with `-llibflatccrt`  and expect this to be a shared library as opposed to `-llibflatccrt_shared`.

@derosier do you have any input on this - having worked on the buildroot distribution?

epouponstormshield:
Hello, I do agree the proposed result is not ideal. Indeed we would expect a libflatccrt.so to be generated!

Valentin-V:
Hi, we now use -DBUILD_SHARED_LIBS=ON, so the patch became useless.
It could be needed if someone has to generate both static and shared lib, but I don't know any use-case for this.

Anyway, thank you for your feedback !

mikkelfj:
OK thanks for the update.

PR #47, opened 2017-06-07
fix compatibility with C++ compilers.

aidancully:
The code generated by flatcc could not be compiled cleanly with
C++ compilers, for the following reasons:

 * C++ does not allow implicit casts from `void *` to
   `destination *`. Addressed by adding explicit type-casts where
   such type-conversions occur, or by adding additional type
   information to source types.
 * C++ deprecates implicit casts from string constants to char *'s,
   or to char array types. Addressed by modifying the constant
   exposed by the `%s_identifier` macros to be of the appropriate
   `flatbuffers_fid_t` data type.

### DIFF
--- a/src/compiler/codegen_c_builder.c
+++ b/src/compiler/codegen_c_builder.c
@@ -170,21 +170,21 @@ int fb_gen_common_c_builder_header(fb_output_t *out)
     fprintf(out->fp,
         "#define __%sbuild_vector_ops(NS, V, N, TN, T)\\\n"
         "static inline T *V ## _extend(NS ## builder_t *B, size_t len)\\\n"
-        "{ return (T *) flatcc_builder_extend_vector(B, len); }\\\n"
+        "{ return (T *)flatcc_builder_extend_vector(B, len); }\\\n"
         "static inline T *V ## _append(NS ## builder_t *B, const T *data, size_t len)\\\n"
-        "{ return (T *) flatcc_builder_append_vector(B, data, len); }\\\n"
+        "{ return (T *)flatcc_builder_append_vector(B, data, len); }\\\n"
         "static inline int V ## _truncate(NS ## builder_t *B, size_t len)\\\n"
         "{ return flatcc_builder_truncate_vector(B, len); }\\\n"
         "static inline T *V ## _edit(NS ## builder_t *B)\\\n"
-        "{ return (T *) flatcc_builder_vector_edit(B); }\\\n"
+        "{ return (T *)flatcc_builder_vector_edit(B); }\\\n"
         "static inline size_t V ## _reserved_len(NS ## builder_t *B)\\\n"
         "{ return flatcc_builder_vector_count(B); }\\\n"
         "static inline T *V ## _push(NS ## builder_t *B, const T *p)\\\n"
-        "{ T *_p; return (_p = (T *) flatcc_builder_extend_vector(B, 1)) ? (memcpy(_p, p, TN ## __size()), _p) : 0; }\\\n"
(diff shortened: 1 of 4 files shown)


mikkelfj:
Thanks. I need to review this more carefully and test against the ci-more branch with older C compilers.

A few comments: you sometimes cast ctx->ctx and sometimes rely on json parser ctx having a new type. Better choose one.

The reason json ctx is void * is to avoid requiring the flatcc builder object in the code being compiled, but I am not sure how careful that design is, and it may also work with a typed (to the application) unknown pointer as long as the pointer is not deferenced. Anyhow, I think I prefer keeping void * there.

aidancully:
Ah, yes, sorry about the inconsistency re: ctx->ctx, I apparently didn't clean things up well enough when trying different approaches. I'll switch `ctx->ctx` back to `void *`, and clean up the rest of the patch, but it will bloat things slightly, since it means a lot more type-casting to `flatcc_builder_t *`.

mikkelfj:
Yes I wondered how much bloat - if it is too bad then lets discuss.

Meanwhile I tested on the older gcc/clang versions, and they were happy.

mikkelfj:
It's OK to change void * to flatcc_builder_t *, the header is included anyway. There are truly many places in the generated code.

aidancully:
OK, since we're sticking with the type-change in `ctx->ctx` to `flatcc_builder_t *`, I've fixed the one other thing you pointed out (unnecessary typecast).

mikkelfj:
I have reviewed most code - looks OK, except a very minor thing - coding style has no space after cast in other code.

I am currently looking into the string cast with fid_t - that's the only thing left.

aidancully:
different projects have different opinions here, for your review, do you prefer I add new commits to the PR? or amend the existing commit?

mikkelfj:
on amend: I don't care - especially since I likely squash.

The fid_t cast is wrong. But it is a bug in flatcc - I think.

I suspect C++ complains about the string cast not because it is a literal, but because fid_t is a modifiable buffer. However, fixing this may require some thought - and I have no capacity left for that today.

mikkelfj:
Perhaps we should make the string issue be a separate commit since it also potentially affects C users. I'd be happy to close the other commits as is. Input welcome.

aidancully:
Fair enough, I'll submit that change in a separate PR. I _think_ the cast is desirable, though we should probably cast to `const flatbuffers_fid_t` rather than to `flatbuffers_fid_t`, but we can discuss in the separate PR.

mikkelfj:
Lets do that. Internally in builder.c the modifiable type is being used, I think. But the defines should be const.

src/compiler/codegen_c_builder.c
```diff
@@ -431,18 +431,18 @@ int fb_gen_common_c_builder_header(fb_output_t *out)
...
+        "{ T *_p; if (!(_p = (T *)flatcc_builder_table_add(B, ID, S, A))) return -1;\\\n"
         "  TN ## _assign_to_pe(_p, v); return 0; }\\\n"
         "\n",
         nsc);
 
     fprintf(out->fp,
         "#define __%sbuild_struct_field(ID, NS, N, TN, S, A)\\\n"
         "static inline TN ## _t *N ## _start(NS ## builder_t *B)\\\n"
-        "{ return flatcc_builder_table_add(B, ID, S, A); }\\\n"
+        "{ return (TN ## _t *) flatcc_builder_table_add(B, ID, S, A); }\\\n"
```
mikkelfj: one space left :)

aidancully: sorry :-)

mikkelfj: noone will know :)

PR #48, opened 2017-06-07
fix const-ness of generated namespace buffer_start and buffer_start_w…

aidancully:
…ith_size functions.

The implementation function expects a `const char` array, and the
the arguments provided to this function are (often) read-only strings.


aidancully:
This also addressed the C++ warnings I was getting. Thanks for the const-ness pointer.

mikkelfj:
So everything works for C++ now?
I still think the generated identifier type is somehow wrong as discussed in #47 - I'll have a look tomorrow.

aidancully:
Yes, with this change and #47, I'm able to compile with `g++ -Wall -Werror` against the generated code.

mikkelfj:
No the identifier is just a define no type involved, so everything looks right. Thanks for fixing this.

Issue #52, opened 2017-07-13
_nest() call unexpected result


AndrewJDR:
Another head scratcher for me.

Given this:
```c
#include <stdio.h>
#include "qmsg.h"

#undef ns
#define ns(x) FLATBUFFERS_WRAP_NAMESPACE(QQQ, x)

int main()
{
  size_t nestedBufSize;
  uint8_t nestedBuffer[1024];

  // Build and store the nested buffer
  {
    flatcc_builder_t pBuilder;
    flatcc_builder_init(&pBuilder);
    ns(Payload_start_as_root(&pBuilder));
      ns(Payload_someId_add(&pBuilder, 456));
    ns(Payload_end_as_root(&pBuilder));

    nestedBufSize = flatcc_builder_get_buffer_size(&pBuilder);
    flatcc_builder_copy_buffer(&pBuilder, nestedBuffer, sizeof(nestedBuffer));
    flatcc_builder_clear(&pBuilder);
  }

  {
    printf("nested buffer: %p nested buffer size:%u\n", nestedBuffer, (uint32_t)nestedBufSize);
    ns(Payload_table_t) nestedPayload = ns(Payload_as_root(nestedBuffer));
    uint32_t someId_before_nesting = ns(Payload_someId(nestedPayload));
    printf("someId (before nesting): %u\n", someId_before_nesting); // <--- prints 456 as expected
  }

  {
    flatcc_builder_t builder;
    flatcc_builder_init(&builder);

    ns(Msg_start_as_root(&builder));
      //ns(Msg_payload_start_as_root(&builder)); <-- also tried with payload_start/end_as_root, prints "12" for someId after nesting.
      ns(Msg_payload_nest(&builder, nestedBuffer, nestedBufSize, 0)); // <--- also tried with 4 and 8 for align. Still prints 0
      //ns(Msg_payload_end_as_root(&builder));
    ns(Msg_end_as_root(&builder));

    size_t bufSize;
    uint8_t *bufPtr = (uint8_t*)flatcc_builder_get_direct_buffer(&builder, &bufSize);
    ns(Msg_table_t) msgTable = ns(Msg_as_root(bufPtr));
    ns(Payload_table_t) nestedPayload = ns(Msg_payload_as_root(msgTable));

    uint32_t someId = ns(Payload_someId(nestedPayload));
    printf("someId (after nesting): %u\n", someId); // <--- Why does this print 0 instead of 456?
  }

  return 0;
}
```

With this schema:
```
[...]

Issue #51, opened 2017-07-12
segfault when using a nested flatbuffer


AndrewJDR:
First off -- this is probably a foolish error on my part.

But here goes. Given:
```c
#include <stdio.h>
#include "qmsg.h"

#undef ns
#define ns(x) FLATBUFFERS_WRAP_NAMESPACE(QQQ, x)

int main()
{
  flatcc_builder_t builder;
  flatcc_builder_init(&builder);

  ns(Msg_start_as_root(&builder));
    ns(Msg_payload_start_as_root(&builder));
      ns(Payload_someId_add(&builder, 456));
    ns(Msg_payload_end_as_root(&builder));
  ns(Msg_end_as_root(&builder));

  size_t bufSize;
  uint8_t *bufPtr = (uint8_t*)flatcc_builder_get_direct_buffer(&builder, &bufSize);
  ns(Msg_table_t) msgTable = ns(Msg_as_root(bufPtr));
  ns(Payload_table_t) nestedPayload = ns(Msg_payload_as_root(msgTable));

  uint32_t someId = ns(Payload_someId(nestedPayload)); // <--- Why a segfault here?
  printf("someId: %u\n", someId);

  return 0;
}
```
With the following schema:
```
namespace QQQ;

table Payload
{
  someId:uint;
}

table Msg
{
  payload:[ubyte] (nested_flatbuffer: "Payload");
}

root_type Msg;
```
I'm segfaulting on the `uint32_t someId = ns(Payload_someId(nestedPayload));` line:
```
bt
#0  0x00000000004008a3 in __flatbuffers_soffset_read_from_pe (p=0x10060b9c0) at flatcc/include/flatcc/flatcc_endian.h:83
#1  0x0000000000400a41 in QQQ_Payload_someId (t=0x10060b9c0) at qmsg.h:943
#2  0x0000000000400fdf in main () at fbs_crash.c:24
```
Would you happen to know why? I've basically tried to follow this example:
https://github.com/dvidelabs/flatcc/blob/master/test/monster_test/monster_test.c#L1736
And as far as I can see, I have done so. Hopefully I'm missing something silly?

Also, my build script (linux):
```
FLATBUFFERS_DIR=../flatcc
FLATCC_EXE=../flatcc/bin/flatcc
FLATCC_INCLUDE=$FLATBUFFERS_DIR/include
FLATCC_LIB=$FLATBUFFERS_DIR/lib

[...]

PR #53, opened 2017-07-13
_nest() was unnecessarily calling _buffer_start()/_buffer_end().

AndrewJDR:
This fixes an issue where an additional table root was inserted just prior
to a _nest()ed flatbuffer, which of course already has its own table root.

### DIFF
--- a/src/compiler/codegen_c_builder.c
+++ b/src/compiler/codegen_c_builder.c
@@ -128,13 +128,11 @@ int fb_gen_common_c_builder_header(fb_output_t *out)
         "static inline int N ## _end_as_typed_root(NS ## builder_t *B)\\\n"
         "{ return N ## _add(B, NS ## buffer_end(B, TN ## _end(B))); }\\\n"
         "static inline int N ## _nest(NS ## builder_t *B, void *data, size_t size, uint16_t align)\\\n"
-        "{ if (NS ## buffer_start(B, FID)) return -1;\\\n"
-        "  return N ## _add(B, NS ## buffer_end(B, flatcc_builder_create_vector(B, data, size, 1,\\\n"
-        "  align ? align : 8, FLATBUFFERS_COUNT_MAX(1)))); }\\\n"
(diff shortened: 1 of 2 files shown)


AndrewJDR:
Attempt to fix issue #52

AndrewJDR:
(Writing test for this currently, will commit it shortly)

mikkelfj:
This is great work.

I wonder if the unwanted create buffer calls helped mask hide #51 where the vector length prefix.

A minor comment on coding style in monster_test.c (as always not arguing about what is right, just what is there):

NULL is tested using (!result) and names are not camelCased and no need to call the test nest_macro, just nest, since depending on view, everything or nothing are macros: nest is actually a function generated by a macro.

Good catch with updating reflection, but if you update this, the version should change in comments, hence the script:

reflection/generate_code.sh

But I tend to run this before releases are as a separate commit to reduce noise.

AndrewJDR:
I think I integrated all of your feedback, let me know if you notice anything missing.
The binary format document was very helpful, I think it's the best reference available anywhere for the flatbuffer format and it helped me understand things well enough to catch this. So thanks again!

> I wonder if the unwanted create buffer calls helped mask #51 where the vector length prefix.

I'm a bit confused on this, how would this misbehavior in _nest() mask issue #51 which was related only to inline nesting? Does inline nesting ultimately use _nest() under the hood? (Doesn't seem to upon cursory glance)

AndrewJDR:
Side note -- I just removed the reflection updates from my commit, so you can just handle that upon your next release whenever that might be.

mikkelfj:
> I'm a bit confused on this, how would this misbehavior in _nest() mask issue #51 which was related only to inline nesting? Does inline nesting ultimately use _nest() under the hood? (Doesn't seem to upon cursory glance)

I'm not sure, just wondering. But I think it is more involved than that. Maybe the code just compiled unexpectedly early and I forgot to add tests.

PR #55, opened 2017-07-25
Adaption for use on embedded devices

frank-freihube-rmcan-com:
By defining what functions to use for malloc/realloc/free in a single place it's much easier to adapt the code to run in embedded environment (e.g. environments using FreeRTOS use pvPortMalloc/vPortFree)
By renaming the include guard it's now less likely that the code conflicts with other libraries as many of them include a portable.h using the same include guard. If a include chain would include both files only the first one would be added.

The functionality of the code should not be affected by the changes made.


mikkelfj:
I'm willing to take a closer look at this - but the current PR will not work
The portably library is independent of flatcc and may be used in other contexts.
Abstractions should be made around portable, for example replacing aligned alloc, or alternatively, adding platforms specific support to portable. For example some RTOS might define it's own aligned_alloc in a sufficiently general fashing that it can be included in portable.

Furthermore, aligned_alloc is very tricky because posix assumes free can be used and this is not always possible, e.g. on windows. So blindly replacing a malloc implementation there can go very wrong.

In other areas, the builder to be exact, the allocator is already customizable, but not at the - replace malloc level. Rather, replace the custom allocator object in the builder. This provides MUCH better control over allocation because you can allocate from a static pool where each object type can have its own space based on assumed usage.

That said, it could make sense to also abstract malloc and free, but do you have a specific use case for this that wouldn't be better off with a custom allocator object?

mikkelfj:
See also flatcc_builder.h

```/**
 * Use instead of `flatcc_builder_init` when providing a custom allocator
 * or emitter. Leave emitter or allocator null to use default.
 * Cleanup of emit and alloc context must be handled manually after
 * the builder is cleared or reset, except if emitter is null the
 * default will be automatically cleared and reset.
 *
 * Returns -1 on failure, 0 on success.
 */
int flatcc_builder_custom_init(flatcc_builder_t *B,
        flatcc_builder_emit_fun *emit, void *emit_context,
        flatcc_builder_alloc_fun *alloc, void *alloc_context);
```

mikkelfj:
As to changing the include guard, that sounds like a good idea but not with FLATCC as a prefix, but an issue separate from malloc/free. I'm considering separating out this library - any suggestions?

frank-freihube-rmcan-com:
Using custom_init is a good idea but I still not understand what the `alloc_context` is for.
Also this will not prevent the library to call `malloc` and `free` in `flatcc_builder_finalize_buffer` or in `flatcc_builder_finalize_aligned_buffer` where `aligned_alloc/_free` defaults to `malloc/free`

frank-freihube-rmcan-com:
Can you give me a hint why AppVeyor fails? I never used MSVC so I'm building the code using an embedded IDE and a Linux VM.

frank-freihube-rmcan-com:
What do you mean by "separating out this library"?

mikkelfj:
> C:\projects\flatcc\src\runtime\builder.c(117): warning C4013: 'flatcc_free' undefined; assuming extern returning int [c:\projects\flatcc\build\MSVC\src\runtime\flatccrt.vcxproj]

flatcc_free is not defined

> What do you mean by "separating out this library"?

The portably library contains 0 features that are specific to flatcc. It aims to support C11 abstractions or some posix abstractions missing or not guaranteed available, including static_assert and aligned_alloc. It also provides unligned_read operations in endian neutral fashion which is incredibly useful in hash functions etc. It is best to keep the library under flatcc but it has more general use so I consider making a separator repository only for this, while keeping a copy here. The portable name is a bit too generic, but very to the point.

> Using custom_init is a good idea but I still not understand what the alloc_context is for.
Also this will not prevent the library to call malloc and free in flatcc_builder_finalize_buffer or in flatcc_builder_finalize_aligned_buffer where aligned_alloc/_free defaults to malloc/free

The alloc_context is (as I recall) just it user define handle. If you alloc memory from addess 0x60000, you might want to tell you custom allocator that the pool starts there and use it as context.

We need to separate the allocation use cases a bit, and there are areas in flatcc that could be improved.

- The builder is the key runtime component that must work everywhere. This is has the custom allocator and it is very performance sensitive. It manages temporary stacks.
- The emitter is just as important, and is also customizable, but the default implementation hardcodes malloc. You could make you own for embedded use.
- Once the buffer is built, the default emitter has a list of pages. This list must be combined to contigous memory, it least for ad hoc use. There are convencience functions specifically for the default allocator which creates a single memory buffer via malloc, or via aligned_alloc if you need more control. But the emitter is roughly just a writev (https://linux.die.net/man/2/writev) function to collect the buffer data being generated so again, for embedded use you will want to control this function, including failing controllably on large buffers since you don't want to check allocating all the time while building buffers (unless autoated via JSON). There is a test where the emitter is changed to a simple logger. Not much code.
- JSON printing need an allocator if using the dynamic buffer, since you cannot predict the output size exactly, but it is entirely optional to use. You can easily add you own printer output, or just print to a fixed size buffer, or to a file. If something is missing it can be added.

So, there aren't really any areas where you don't have full control. My thinking is that if you system really don't want to use malloc, then just redefine malloc and free. But I am willing to consider making it configurable via include/flatcc/flatcc_config.h - I just don't want to overengineer this - since - likely you want to customize the emitter and allocator anyway on such low-level systems, and overriding aligned_alloc is non-trivial. https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_rtconfig.h

Perhaps the issue here is the need for a separate documentation section on this?
Aside from the include guard thing.

mikkelfj:
On the Windows build issue:

portable.h is not included by default. You need to define  FLATCC_PORTABLE to achieve this. Some old compilers need it. For general include management in flatcc, the files of interest are include/flatcc_flatbuffers.h and include/flatcc_portable.h. This is where you should override aligned_alloc if necessary, rather than modifying paligned_alloc.h. You would then `#define __aligned_free_is_defined 1` to disable the portable library version. If that doesn't work, there is a bug in paligned_alloc.h

mikkelfj:
I have fixed the include guard issue in cc73100b98d4ac28318346e4a5d3e303ccadc05a
and 59936b0f9eff88745048d2648188020c5adc95b3

mikkelfj:
1a780fb012cb765e48df38705a52979c9ac9ab70
fixes flatcc_emitter.h - it should use aligned_free, an at the same time I added support for overriding free.

mikkelfj:
Added support for overriding the default allocator in the builder
 11d43b62a5f088fc1cc7c5fcd9be158a2f19f7c4

mikkelfj:
I have now roughly reimplemented you PR without touching portable library, except for fixing the include guard issue.

Please let me know if this addresses you concerns.

You still need to add a custom aligned_alloc, either by properly extending paligned_alloc.h with a new known system, or by defining you own in an early include file, using __aligned_alloc_is_defined and __aligned_free_is_defined to disable the default implementation on a given platform.

mikkelfj:
Perhaps I should add a common FLATCC_ALLOC as well?

mikkelfj:
Now there is a flatcc/flatcc_alloc.h file to control allocation
a2de05fd813ff2a5c93874b2a71288cc998e6820

frank-freihube-rmcan-com:
I think I understand now what's meant with the context. If I get things right this is only for user specific support and does nothing in the default code.

frank-freihube-rmcan-com:
I do not mind using hardcoding for performance reasons, therefor I used the preprocessor replacement.
I like what you did in `flatcc_alloc.h` but problem is, that `paligned_alloc.h` still uses `malloc` and `free` which still have to be replaced.
what would you think about adding `palloc.h` only defining what function to use for mallod/realloc/free and including them in `paligned_alloc.h`?
This way one has to change one file only to replace the functions wherever needed and one would change portable which makes more sense then changing flatcc
Thing is in our context malloc and free exist but may not be called because FreeRTOS brings their own functions which do not use controllers heap. Calls to malloc would fail because heap is set to size 0.
I reverted my changes, apllied yours and added what I described above. I hope you can agree to that.

frank-freihube-rmcan-com:
To extract portable as a library I would use git submodules instead of a static copy. This way its easier to keep them consistent and be able to switch versions.
We are working with git submodules a lot and it works great for us.

mikkelfj:
Hi Frank, thanks for the input.
I understand what you are trying to achieve but it mixes it different concepts:
The purposes of the portable library is to allow applications to write posix C11 without worrying. Redefinig malloc would be wrong in this context even if it is usable.

But, there is a use case for helping platforms that cannot use malloc for some reason. But patching aligned_alloc is not that use case.

We could add a palloc.h file but to make sense of that, there would need to be some concrete platforms to be detected and it would not pollute the implementation of aligned_alloc. Rather it would provide its own alternative implementation. Since I don't have sufficient material to add such a portable alloc, flatcc_alloc.h is the next best thing. So this is where an alternative aligned alloc could be implemented if needed. But we would still need some configuration flags to test which version to use.

As to git submodules: I don't think there are any simple solutions to this problem. Using submodules forces dependency on a server location, here GitHub, and a specific build system. The complexity is not worth the cost of copying a specific known version of a library. For example Homebrew package building would need an extra build step for no apparent benefit.

mikkelfj:
I added a flatcc_alloc.h implementation of FLATCC_ALIGNED_ALLOC
787c7dd9d024b3b3812a2fb7082e7c7e0244c270

If we have proper system tests to detect FreeRTOS, we could move parts of flatcc_alloc.h into palloc.h
and define portable_malloc, portable_free, portable_aligned_alloc etc. Then make flatcc_alloc use those. I don't think pmalloc would be good since it could easily conflict. pMalloc would go against naming conventions already in use.

frank-freihube-rmcan-com:
Now I understand what you were trying to do. Works fine now.
Thanks to your last changes I do not even have to change anything in flatcc, it's enough if I define `FLATCC_ALLOC/_FREE/_REALLOC_CEALLOC` first before including anything from flatcc.
Thank you very much for your time.
One last thing, you added `FLATCC_CALLOC` but as far as I can tell it's not used. Is it just for the sake of completeness?

BTW: there are two typos in `flatcc_alloc.h`:
* in `__flatcc_aligned_alloc` : "FLACC_ALLOC" (missing "T")
* in `#define FLATCC_ALIGNED_ALLOC` alignment != alignemnt

mikkelfj:
> BTW: there are two typos in flatcc_alloc.h:

I know I didn't test the last changes behind special flags properly, hence typo. Could you please update the PR to fix these since you apparently have a going setup.

> One last thing, you added FLATCC_CALLOC but as far as I can tell it's not used. Is it just for the sake of completeness?

Yes in part - I noticed I had calloc calls in source, but then realized they were only used with the compiler, not runtime. But I kept it for completeness.

Note that you also have to define FLATCC_USE_GENERIC_ALIGNED_ALLOC=1
unless you define you own aligned alloc version. But I guess you have figured that out.

frank-freihube-rmcan-com:
Did it, but maybe you want to extract the changes and close the PR as there are many revert commits in it.
Also I do not use this code too, just recognised this by examine the code to understand what you were doing ;)

mikkelfj:
Thanks, I'm squashing the merge so no issue with reversals.

frank-freihube-rmcan-com:
No, did not figured that out. I were not defining this but it still worked. But I removed all `aligned_free` from `monster.c` (technically no library code but application code) and used FreeRTOSs `vPortFree` instead.
Doesn't the library use the `flatcc_alloc`-Functions?

mikkelfj:
Uh, the library does, but monster.c is a a user application. If you override Posix allocations you also need to do it in user applications - I don't think it would be helpful to place FLATCC_ALIGNED_FREE in examples.

But, this is a perfect example of why it is better to implement and use Posix rather than some special library specific functions. I don't really like doing it in flatcc, but I see it's use since some platforms are just difficult to work with.

mikkelfj:
Clarification: libflatccrt use flatcc_alloc.h macros, the libflatcc compiler library does not, except those part that include the runtime library - so one should be a bit carefuly about mixing this. I need to add a comment in flatcc_alloc.h

PR #59, opened 2017-10-01
Fixes a couple of issues when building with c++.

AndrewJDR:
When using flatcc to generate code that will be used by a c++ compiler (clang++ in my case), errors like this will be encountered on struct types:
```
fatal error: cannot initialize return object of type 'SomeType *' (aka 'SomeType *') with an rvalue of type 'void *'
```

This fixes the issue using typecasts.


mikkelfj:
Thanks Andrew, looks reasonable - we havde done similar changes before I believe. I just wonder why wrapping in extern C isn't enough, or was that not practical? Or are we missing extern C in generated code? We risk going down a rabbit hole of ever changing C++ standards... opinion?

We ought to have an automated build on Travis and AppVeyor with a C++ configuration, and a CMake flag, but that is one of those corners I have cut.

AndrewJDR:
I was using extern "C" around the generated header. I don't know why that wasn't enough, to be honest. I could dig more into it to find out why, but it'll have to be a bit later... in the middle of a project this week.

mikkelfj:
I think someone else came to a similar conclusion - if you find something, it would be good to know, otherwise we can just add casts. In general though, it is a feature of C to be able to return void pointers, unsafe as it is, so I dislike C code being forced to do this because of C++.

PR #69, opened 2018-01-02
C++11 compatibility

aidancully:


mikkelfj:
Is this sufficient? The following specializes the checks on different compilers for C++

https://github.com/wc-duck/dbgtools/blob/master/include/dbgtools/static_assert.h

aidancully:
I based that on http://en.cppreference.com/w/cpp/language/static_assert (describing that the keyword was introduced in C++11) and http://en.cppreference.com/w/cpp/preprocessor/replace (see the explanation of the __cplusplus macro), under the assumption that the cppreference site was fairly authoritative about the standard. I suspect that the compiler-specific approach will work with more compilers than the standards-based approach I took here, but I also think the standards-based approach should be sufficient.

mikkelfj:
It may well be that your PR is sufficient, but it is hard to tell - compilers and clib implementations are not always fully compliant.

Recent std C uses a trick with __xxx_is_defined to handle xxx as a keyword. I have used this elsewhere, and added it to static_assert along with a more robust version in branch "static_assert".

Please review and possibly include in PR.

https://github.com/dvidelabs/flatcc/blob/static_assert/include/flatcc/portable/pstatic_assert.h

mikkelfj:
I revised a few times. I forgot that _Static_assert might be defined in C11 while the clib headers don't. This is the case for clang on MacOS, or at least used to be the case, so might be true for other compilers as well.

Last revision here:
https://github.com/dvidelabs/flatcc/commit/5fa73a34b69b0abdd3607d0319d756ea82cdb9d1

aidancully:
OK, I've incorporated your version in the PR, thanks!

PR #70, opened 2018-01-06
Gitmask Anonymous PR

gitmask-anonymous:
This is an anonymous PR submitted via Gitmask - https://www.gitmask.com

### DIFF
--- a/src/runtime/json_printer.c
+++ b/src/runtime/json_printer.c
@@ -167,10 +167,10 @@ static void print_string_part(flatcc_json_printer_t *ctx, const char *name, size
 {
     size_t k;
 
+    if (ctx->p >= ctx->pflush) {
+        ctx->flush(ctx, 0);
+    }
     if (ctx->p + len >= ctx->pflush) {
-        if (ctx->p >= ctx->pflush) {
-            ctx->flush(ctx, 0);
-        }
         do {
             k = ctx->pflush - ctx->p;
             memcpy(ctx->p, name, k);


mikkelfj:
While I appreciate user contributions, it's a bit hard to have dialog this way.
If you do not wish to comment here or raise an issue via a GitHub user, at least supply a patch with a comment motivating the change. You can also email me privately at mikkel@dvide.com.

Is there a bug or performance issue in the original code that I am not aware of?

The patch requires two if conditions for every string part while the original code avoids an extra check when printing small strings into a large print buffer.

The original code has a seemingly redundant if condition which is present to avoid a negative k value. This can happen because p can exceed pflush by a small limited amount when printing integers or other small values.

The patch correctly keeps this check that protects against overflow, but apparently with the only effect of slowing down the average case.

prcomments:
Sorry for the inconvenience. The motivation behind the change is to mitigate the error occurring when ctx->p is already at ctx->pflush, and print_string_part is called. What happens then is the following:

- the current content gets flushed,
- ctx->p gets reset to ctx->buf,
- k, computed as  ctx->pflush - ctx->p, becomes the same as ctx->flush_size, and
- memcpy(ctx->p, name, k) reads ctx->flush_size bytes at name, but there are len bytes available and len is usually smaller than ctx->flush_size.

I see now that the patch slows down the average case, and that can be avoided. One way to do so is to keep the conditional flush inside the ctx->p + len >= ctx->pflush condition, but check that same condition again after the flush. What do you think?

Thank you

mikkelfj:
Thanks for explaining and reporting this, that is indeed unfortunate. I see `print_indent_ex` has a similar solution. I made two commits, one doing a minimal change to fix the issue b55220a2a68dc144f7072c555e6306dd3e6b624b, and a second one that cleans up naming and speeds up printing slightly by inlining the fast path similar to how indent works 1b8a4dbe01ebf3f322ada97b8e34b0af43b4a44c

Can you please confirm this resolves the issue?

I move benchmark printing from 516 - 520MB/s to about 520 - 536MB/s with the second commit.

test/benchmark/benchflatccjson/run.sh

mikkelfj:
122f175fd8d1353f4d5705097eb3ad1d0ba7a162 fixes the most recent commit because ctx->flush_size is not the capacity after a flush due to spill-over being moved down. The first minimal fix does not have this issue.

mikkelfj:
8f5bfd2fb0a6e3db3e9f2ca84759fecf284bd16e

prcomments:
https://github.com/dvidelabs/flatcc/commit/b55220a2a68dc144f7072c555e6306dd3e6b624b resolves the issue I had with print_string_part, thank you!

PR #71, opened 2018-01-09
Add cast to (char *) to avoid -fpermissive error when using C++

paulreimer:
Avoid the following error when using flatcc from C++ via this PR:
```
In file included from /X/flatcc/include/flatcc/portable/pstdalign.h:60:0,
                 from /X/flatcc/include/flatcc/flatcc_flatbuffers.h:20,
                 from /X/flatcc/include/flatcc/flatcc_builder.h:64,
                 from /X/src/json_to_flatbuffers_converter.inl:9,
                 from /X/src/json_to_flatbuffers_converter.h:31,
                 from /X/curl_test_task.cpp:3:
/X/flatcc/include/flatcc/portable/paligned_alloc.h: In function 'void __portable_aligned_free(void*)':
/X/flatcc/include/flatcc/portable/paligned_alloc.h:131:32: error: invalid conversion from 'void*' to 'char*' [-fpermi]
     char *raw = ((void **)p)[-1];
```

### DIFF
--- a/include/flatcc/portable/paligned_alloc.h
+++ b/include/flatcc/portable/paligned_alloc.h
@@ -128,7 +128,7 @@ static inline void *__portable_aligned_alloc(size_t alignment, size_t size)
 
 static inline void __portable_aligned_free(void *p)
 {
-    char *raw = ((void **)p)[-1];
+    char *raw = (char*)((void **)p)[-1];
 
     free(raw);
 }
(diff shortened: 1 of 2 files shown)


paulreimer:
Commit 50adbbd addresses the following error:
```
In file included from /X/flatcc/src/runtime/json_printer.c:26:0:
/X/flatcc/include/flatcc/portable/pbase64.h: In function 'base64_encode':
/X/flatcc/include/flatcc/portable/pbase64.h:221:18: error: 'len' may be used uninitialized in this function [-Werror=]
         *src_len -= len;
 ```

mikkelfj:
Thanks. Just for the record, in pbase64.h the `len`variable is guarded so the runtime is not affected, but it is still cleaner to have it initialized.

PR #72, opened 2018-01-09
[WIP] Add '#ifdef __cplusplus' guards 

paulreimer:
Wrap public header files in `#ifdef _cplusplus` guards to allow `#include`s from other C++ or C translation units.

Also add support to codegen_c sources to add include guards to all generated headers.

This helps me when using flatcc from C++, as I do not need to wrap the flatcc header includes in `extern "C"`.

As a first attempt, I added it manually to (almost all) header files in `include`, just after the existing `#ifdef FILE_H` include guards (if present). This generates files such as I modify them by hand, and for my purposes appears to work fine.

### DIFF
--- a/include/flatcc/flatcc.h
+++ b/include/flatcc/flatcc.h
@@ -1,6 +1,10 @@
 #ifndef FLATCC_H
 #define FLATCC_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /*
  * This is the primary `flatcc` interface when compiling `flatcc` as a
  * library. Functions and types in the this interface will be kept
@@ -250,6 +254,10 @@ void flatcc_destroy_context(flatcc_context_t ctx);
 
 #ifdef _MSC_VER
 #pragma warning(pop)
+#endif
+
+#ifdef __cplusplus
+}
 #endif
 
 #endif /* FLATCC_H */
--- a/include/flatcc/flatcc_accessors.h
+++ b/include/flatcc/flatcc_accessors.h
@@ -1,6 +1,10 @@
 #ifndef FLATCC_ACCESSORS
 #define FLATCC_ACCESSORS
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #ifndef UINT8_MAX
 #include <stdint.h>
 #endif
@@ -85,5 +89,9 @@ __flatcc_define_basic_integer_accessors(NS, int32, int32_t, 32, E)          \
 __flatcc_define_basic_integer_accessors(NS, int64, int64_t, 64, E)          \
 __flatcc_define_basic_real_accessors(NS, float, float, 32, E)               \
 __flatcc_define_basic_real_accessors(NS, double, double, 64, E)
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif /* FLATCC_ACCESSORS */
--- a/include/flatcc/flatcc_alloc.h
+++ b/include/flatcc/flatcc_alloc.h
@@ -1,6 +1,10 @@
 #ifndef FLATCC_ALLOC_H
 #define FLATCC_ALLOC_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /*
  * These allocation abstractions are __only__ for runtime libraries.
  *
@@ -111,6 +115,10 @@ static inline void __flatcc_aligned_free(void *p)
 
 #ifndef FLATCC_ALIGNED_FREE
 #define FLATCC_ALIGNED_FREE(p) aligned_free(p)
+#endif
+
+#ifdef __cplusplus
+}
 #endif
 
 #endif /* FLATCC_ALLOC_H */
--- a/include/flatcc/flatcc_builder.h
+++ b/include/flatcc/flatcc_builder.h
@@ -1,6 +1,10 @@
 #ifndef FLATCC_BUILDER_H
 #define FLATCC_BUILDER_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /**
  * Library for building untyped FlatBuffers. Intended as a support
  * library for generated C code to produce typed builders, but might
@@ -1754,5 +1758,9 @@ void flatcc_builder_aligned_free(void *p);
  * Other emitters have custom interfaces for reaching their content.
  */
 void *flatcc_builder_copy_buffer(flatcc_builder_t *B, void *buffer, size_t size);
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif /* FLATCC_BUILDER_H */
--- a/include/flatcc/flatcc_emitter.h
+++ b/include/flatcc/flatcc_emitter.h
@@ -1,6 +1,10 @@
 #ifndef FLATCC_EMITTER_H
 #define FLATCC_EMITTER_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /*
  * Default implementation of a flatbuilder emitter.
  *
@@ -203,5 +207,9 @@ void *flatcc_emitter_copy_buffer(flatcc_emitter_t *E, void *buf, size_t size);
 int flatcc_emitter(void *emit_context,
         const flatcc_iovec_t *iov, int iov_count,
         flatbuffers_soffset_t offset, size_t len);
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif /* FLATCC_EMITTER_H */
--- a/include/flatcc/flatcc_endian.h
+++ b/include/flatcc/flatcc_endian.h
@@ -1,6 +1,10 @@
 #ifndef FLATCC_ENDIAN_H
 #define FLATCC_ENDIAN_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /*
  * This file provides helper macros to define type-specific macros and
  * inline functions that convert between stored data and native data
@@ -112,6 +116,10 @@
 
 #ifndef flatbuffers_is_native_be
 #define flatbuffers_is_native_be() (!flatbuffers_is_native_pe())
+#endif
+
+#ifdef __cplusplus
+}
 #endif
 
 #endif /* FLATCC_ENDIAN_H */
--- a/include/flatcc/flatcc_flatbuffers.h
+++ b/include/flatcc/flatcc_flatbuffers.h
@@ -9,6 +9,10 @@
 #ifndef FLATCC_FLATBUFFERS_H
 #define FLATCC_FLATBUFFERS_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #ifndef flatcc_flatbuffers_defined
 #define flatcc_flatbuffers_defined
 
@@ -39,5 +43,9 @@
 #endif
 
 #endif /* flatcc_flatbuffers_defined */
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif /* FLATCC_FLATBUFFERS_H */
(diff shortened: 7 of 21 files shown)


mikkelfj:
How about flatcc/portable/pdiagnostic_push/pop. It's there to deal with compiler specifics.

Note that it is not flatcc specific. I'm not 100% convinced, but it is included to avoid a ton of specialization in generated code. It should also guard the static header files if include order is correct.

https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pdiagnostic_push.h

#if defined(__cplusplus) && !defined(PDIAGNOSTIC_NO_CPLUSPLUS_GUARD)
extern "C" {
#if

https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pdiagnostic_pop.h 

#if defined(__cplusplus) && !defined(PDIAGNOSTIC_NO_CPLUSPLUS_GUARD)
} /* extern "C"
#endif

mikkelfj:
I periodically get these small patches to C++. If a small C++ test project was created under tests, it would capture most of these issues so if you or anyone else up for the task, it would be appreciated. It should build on AppVeyor and Travis.

mikkelfj:
I edited the original pdiagnostic proposal to avoid nested extern "C". Nesting should work on most compilers, but there is no reason fail on the odd compiler that does not.

paulreimer:
What about convincing CMake to use the C++ compiler for top-level test c files, as a second pass on the same tests?

Also, I can split the PR into patching the codegen, vs patching the existing headers (I don't use portable, so I wouldn't have caught the above). In my case, use of flatcc is already scoped `extern "C"`'d in a library, but application code still needs to include the generated headers.

paulreimer:
Looking at the `test/` layout, I think you are right that it begs a `cpp_test` subfolder.

mikkelfj:
A part of portable is always used to a degree in order to handle endian detection and unaligned reads etc., and to deal with compiler specific warnings.

I'm not sure I like the idea of rerunning CMake with a separate compiler for all tests. There is the ci-more branch that drives a lot of different C compilers already via Travis and Appveyor. I merge into this branch close to a major release.

I would rather avoid these guards all over the code generator if it can be avoided, so I am not too keen on the PR in its current form. A lot of work has been made to move logic into a system of include files. The early version of flatcc had some pretty messy output.

It might be an idea to have guards in all the include/flatcc files because you are not always necessarily including via generated files. This has already been done to all the portable files (I suppose).

Regardless, I'd like to see if the pdiagnostic idea makes any sense, both in practical terms and if it makes sense to you.

I probably need a bit of time to consider the options but I will be releasing 0.5.1 soon due to a buffer overrun discovered in the JSON printer.

paulreimer:
Agree with you about keeping it out of the codegen, if everything generated should be fine when compiled as C++ (and every `#include`d would be guarded within), then I can revert 3400280.

If I catch your idea about putting the guard in `pdiagnostic_push.h` / `pdiagnostic_pop.h`, then the generated code is already emitting this as the first/last `#include` so that is a good place to put guards? In that case, I recommend those files be renamed -- not sure what is best -- to reflect that purpose (where it was once diagnostics-specific and made sense to be called `pdiagnostic_*`, now there are multiple responsibilities for that file).

I have usually done it the boring way and copy/pasted it in all the public headers, and when I am reading through headers I know to be for a C-based library, it is a quick mental checkmark to see the `#ifdef __cplusplus`. But ultimately I just want to resolve the right symbols, however that is best.

As to guarding the flatcc headers, I thought all the user-facing include/flatcc headers, as well as include/flatcc/portable, were guarded in db26c34, if that's what you mean?

paulreimer:
`required_macros_push.h` / `_pop.h`?

mikkelfj:
> If I catch your idea about putting the guard 

You did catch it.

> I recommend those files be renamed

Yes, that has occurred to me too. This is one reason I need to ponder a bit. But it doesn't have to be done all at once. It could also be a new file named pprologue.h and pepilogue.h which in turn include the pdiagnostic push/pop and set the C++ guard.

>  it is a quick mental checkmark to see the #ifdef __cplusplus. 
Yes, that is a good point - this is also a reason I need to ponder a bit. After all there is only one __cplusplus, unlike compiler warnings in general but it is also convenient that the generated output is uncluttered.

> As to guarding the flatcc headers, I thought all the user-facing include/flatcc headers, as well as include/flatcc/portable, were guarded in db26c34, if that's what you mean?

Yes, I meant to say that the part of the PR might be a good idea though not stricly necessary. I have not closely reviewed all details, but that was my impression.

I suggest I get some much needed sleep before making any decisions and you experiment a bit, and we can see what makes sense tomorrow.

mikkelfj:
> required_macros_push.h / _pop.h?

I think pprologue.h and pepilogue.h are better. All portable files are prefixed p, in continuation of pstdint.h, and to avoid easy name conflicts (portable.h of course, was a mistake).

paulreimer:
Gotcha. I might make that patch and see how it goes. Also I can just copy the monster test, I think, and make it .cpp? I don't think it has to do anything fancy, just the same code with a C++ compiler. Maybe a CMake option to avoid a build dependency on a C++ compiler.

mikkelfj:
If you copy the monster test, it should be done as an include similar to how other tests reference various files in the monster_test directory. An additional or alternative test could be some actual C++ code that does a very small buffer project with one of the tables in the monster schema. Ideally json parsing and printing should also be covered.

WARNING: I am about to push some non-trivial changes to monster_test.fbs and monster_test.c in order to align better with the C++ version of monster_test.fbs. It will happen likely happen tomorrow as it is late here.

WARNING2: I have significantly reorganized the JSON parsing test to be more modular and allow testing of arbitrary tables with the same test driver. These changes are also likely to arrive tomorrow.

paulreimer:
So... no testing right now and only implementation?

No problem :wink:

mikkelfj:
To clarify: you cannot just copy and rename because monster_test.c is changing frequently and is partially also used as rough documentation and reference material. It would get out of date quickly and be difficult to maintain. This is why it should be included if that approach is taken.

> Maybe a CMake option to avoid a build dependency on a C++ compiler.

Yes, that could be a good idea. I have actually also contemplated that at a high level. There are guards in CMake to exclude tests already, so you can probably isolate C++ to the test environment last in the main CMake file or in the test directory.

mikkelfj:
I landed the changes now. I lost a few percent of performance in the json benchmark due to code generation cleanup but I can address that later.

I always run scripts/test.sh before committing anything signicant.

mikkelfj:
Reason to avoid nesting extern "C:
https://sourceforge.net/p/asn1c/bugs/14/
https://gcc.gnu.org/ml/gcc-prs/2001-10/msg00712.html

mikkelfj:
Some thoughts after thinking about this a bit. Perhaps it is not the best idea to add guards in pdiagnostic or pprologue. If there is a new pprologue file it should probably by include/flatcc/flatcc_prologue.h, include/flatcc_epilogue.h which then include diagnostics.

Some review comments:

Changes to reflection makes it hard to review. Since these changes do not break builds (at least not without a C++ test case), they are not needed. For details please refer to the updated README:
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/reflection/README.md

pstdint.h is important from  Paul Hsieh it I'd like to keep that file unchanged so that it can be updated as needed. It is only included in a few places and these could be guarded by __cplusplus checks.

Some files already handle __cplusplus such as pstdbool.h so perhaps a bit more careful analysis on where guards are needed. pbase64.h is a new file, but otherwise I think I already went through the files in portable, although I might have missed something.

When adding __cplusplus guards it is best to add them after including other files to avoid nesting or forcing them into extern C if they indeed need a C++ section. This is true for both generated and library code.

But other than that - I am still considering the options and awaiting your input.

mikkelfj:
Actually, pdiagnostics have long bothered me, so I think it is a good idea with flatcc_epilogue.h etc. regardless of whether __cplusplus guards are placed there or not.

mikkelfj:
Added a prologue branch with codegen guards
https://github.com/dvidelabs/flatcc/commit/db6bf640c1233bdefa0af64a4c45d226e7616925

paulreimer:
I like the changes in that branch, the naming (`flatcc_prologue`) seems better than `portable/pprologue`. Also, that branch removes the need to wrap the generated headers with `extern "C"` when they are included from C++ user code. So, awesome! I can rebase a new PR on that, and add the changes from the first commit in this PR 2ebc67c (i.e. copy/pasting `#ifdef __cplusplus` on `include/*.h`).

paulreimer:
I might even say that branch is sufficient and this PR is not needed, since in my library I just use `extern "C"` one time around the `flatcc/` headers and it's done.

mikkelfj:
Well then let's leave it at that for now - I'll merge prologue branch as I'd like to get a 0.5.1 release out. If you wan't to polish you are welcome to submit a separate PR for include/flatcc/flatc_*.h and include/flatcc/portable/*.h based on review comments - eventually it probably has to be done, I just didn't bother because there are not tests covering it anyway. A cpp test projects is probably more important.

paulreimer:
OK sounds good. Hope you notice PR #73 also, it prevents me from compiling the generated JSON headers with a C++ compiler. Other than that, 0.5.1 sounds like it'll solve nearly all my issues.

paulreimer:
FYI I rebased this on latest `master` and removed the codegen changes. So this PR now represents just adding `__cplusplus` guards to all headers within `include/` (naively / nested).

mikkelfj:
OK, but I won't merge this as is, referring to previous review comments. Notably pstdint and reflection.

mikkelfj:
I just went through all the changes to `flatcc_*`. I think it is OK to merge these - it gets messy to try to place under include directives in general, and it also covers all the portable files.

Portable therefore does not need the wrappers unless a C++ file is explicitly referencing a portable file, and that shouldn't be needed.

grisu3 and pbase64 probably should have gaurds, but most others probably should not, so that can be fixed separately. Also, grisu3 must be sync'ed with the external directory that has the full source with tests.

So I suggest you modify you commit to only include the `flatcc_*` files and test that this is indeed sufficient to avoid extern declarations in your C++ source. If so, I can merge this into 0.5.1

mikkelfj:
I didn't look at support, but I guess these are also ok to include - they aren't magic.

mikkelfj:
I went through all the portable files and added wrappers where it appeared to make sense.

https://github.com/dvidelabs/flatcc/commit/fa81fd95751e8db3ab121b20d2acb947ef5e032f

You can include that commit if you like.

mikkelfj:
Without reflection

paulreimer:
Alright, I reverted the `portable` header changes, and now also removed `reflection`. `support` changes are still in there. Also, I should note that I think you recommended moving the include guards *after* each file's own `#include`s, and I haven't done that.

mikkelfj:
Thanks. On include order - I reverted on that - as I wrote somewhere in the last hour or so. The reason being it got quite messy with the cplus thing quite far down inside some files. I don't think avoiding  covering includes was important enough for that.

Have you notice my portable commit with those changes. If you test with those, I can merge them separately.

paulreimer:
Sure, I merged those in to a local branch based on this PR, and it works.

I still get the `pstdalign.h:36:0: warning: "_Alignas" redefined` warning, and that warning is emitted when compiling the `src/runtime/*.c` files as well (which I did not put the workaround you posted earlier)

paulreimer:
Also I should note that I no longer require `extern "C"` anywhere, with only the changes in this PR (but the other branch changes seem worthwhile, too)

mikkelfj:
That is fine. Lets deal with pstdalign separately in other PR or a new if you create one.

PR #73, opened 2018-01-11
Fix incorrect void * in generated _as_root JSON parsers

paulreimer:
to support compiling the json_parser header with a C++ compiler.

### DIFF
--- a/src/compiler/codegen_c_json_parser.c
+++ b/src/compiler/codegen_c_json_parser.c
@@ -1477,7 +1477,7 @@ static int gen_table_parser(fb_output_t *out, fb_compound_type_t *ct)
     println(out, "");
     println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const void *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
     println(out, "{"); indent();
-    println(out, "return flatcc_json_parser_table_as_root(B, ctx, buf, bufsiz, flags, fid, %s_parse_json_table);",
+    println(out, "return flatcc_json_parser_table_as_root(B, ctx, (char*)buf, bufsiz, flags, fid, %s_parse_json_table);",
             snt.text);
     unindent(); println(out, "}");
     println(out, "");
--- a/src/compiler/codegen_c_json_parser.c
+++ b/src/compiler/codegen_c_json_parser.c
@@ -1371,7 +1371,7 @@ static int gen_struct_parser(fb_output_t *out, fb_compound_type_t *ct)
     println(out, "");
     /* TODO: */
 #if 0
-    println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const void *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
+    println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const char *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
     println(out, "{"); indent();
     println(out, "return flatcc_json_parser_struct_as_root(B, ctx, buf, bufsiz, flags, fid, %s_parse_json_struct);",
             snt.text);
@@ -1475,9 +1475,9 @@ static int gen_table_parser(fb_output_t *out, fb_compound_type_t *ct)
     println(out, "return flatcc_json_parser_set_error(ctx, buf, end, flatcc_json_parser_error_runtime);");
     unindent(); println(out, "}");
     println(out, "");
-    println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const void *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
+    println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const char *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
     println(out, "{"); indent();
-    println(out, "return flatcc_json_parser_table_as_root(B, ctx, (char*)buf, bufsiz, flags, fid, %s_parse_json_table);",
+    println(out, "return flatcc_json_parser_table_as_root(B, ctx, buf, bufsiz, flags, fid, %s_parse_json_table);",
             snt.text);
     unindent(); println(out, "}");
     println(out, "");


paulreimer:
Fixes this error:
```
In file included from ./curl_test_task.cpp:8:0:
./user_test_json_parser.h: In function 'int UserTest_User_parse_json_as_root(flatcc_builder_t*, flatcc_json_parser_t*, const void*, size_t, int, const char*)':
./user_test_json_parser.h:134:108: error: invalid conversion from 'const void*' to 'const char*' [-fpermissive]
     return flatcc_json_parser_table_as_root(B, ctx, buf, bufsiz, flags, fid, UserTest_User_parse_json_table);
                                                                                                            ^
In file included from ./src/json_to_flatbuffers_converter.inl:10:0,
                 from ./src/json_to_flatbuffers_converter.h:31,
                 from ./curl_test_task.cpp:3:
./flatcc/include/flatcc/flatcc_json_parser.h:866:5: note:   initializing argument 3 of 'int flatcc_json_parser_table_as_root(flatcc_builder_t*, flatcc_json_parser_t*, const char*, size_t, int, const char*, const char* (*)(flatcc_json_parser_t*, const char*, const char*, flatcc_builder_ref_t*))'
 int flatcc_json_parser_table_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx,
     ^
```

src/compiler/codegen_c_json_parser.c
```diff
@@ -1477,7 +1477,7 @@ static int gen_table_parser(fb_output_t *out, fb_compound_type_t *ct)
     println(out, "");
     println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const void *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
     println(out, "{"); indent();
-    println(out, "return flatcc_json_parser_table_as_root(B, ctx, buf, bufsiz, flags, fid, %s_parse_json_table);",
+    println(out, "return flatcc_json_parser_table_as_root(B, ctx, (char*)buf, bufsiz, flags, fid, %s_parse_json_table);",
             snt.text);
```
mikkelfj: It's the `buf` argument that is wrong. It should be const char * not const void *. Note that the prototype is already generated earlier in the file.

paulreimer:
Alright, I think you mean that 08ddbf7 is safe. However I think it could break other code that relies on the old function signature from headers generated by a previous version. That doesn't apply to me, and personally I think it is better as `char*` anyway as you suggested.

src/compiler/codegen_c_json_parser.c
mikkelfj: Right. But there was no prototype, my mistake, nor is one needed. But you found the dead code for the struct parser. It should actually be enabled. I'll merge this and figure it out why the code was dead. I'll be good if you can test again when I get to it, I'm on it.

src/compiler/codegen_c_json_parser.c
```diff
@@ -1371,7 +1371,7 @@ static int gen_struct_parser(fb_output_t *out, fb_compound_type_t *ct)
     println(out, "");
     /* TODO: */
 #if 0
-    println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const void *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
+    println(out, "static inline int %s_parse_json_as_root(flatcc_builder_t *B, flatcc_json_parser_t *ctx, const char *buf, size_t bufsiz, int flags, const char *fid)", snt.text);
```
paulreimer: I think you were describing this declaration but seems it doesn't isn't used at the moment.

mikkelfj: yes - see other comment

mikkelfj:
>Alright, I think you mean that 08ddbf7 is safe. However I think it could break other code that relies on the old function signature from headers generated by a previous version. That doesn't apply to me, and personally I think it is better as char* anyway as you suggested.

There has been some breaking changes to 0.5.0, and 0.5.1 follows shortly. I'd rather take the pain now. Argument could also have been uint8_t *, but const char * is used systematically elsewhere (I hope).

paulreimer:
The use of `char*` makes it just a bit easier for me to use C++ `string_view`'s, which is nice.

mikkelfj:
OK, please test again. Apparent I had two table_as_root library function prototypes in parser.h instead of struct and table thing, that broke the build. Should be ok now. Please test.

mikkelfj:
> The use of char* makes it just a bit easier for me to use C++ string_views, which is nice.

Right. In many other api's the trend is uint8 when dealing with arbitrary data such as base64 and encryption, but here char makes sense.

paulreimer:
It appears to (still) work for me, with the latest master (ccd16b5). Great! There is one warning remaining when compiling with C++, but I will make a separate issue for reporting that.

mikkelfj:
We need that test case :-) It could be the sample/monster project, it doesn't change because it is part of the tutorial.

mikkelfj:
Just waiting for your new issue, before releasing 0.5.1

paulreimer:
Sorry for the delay, I had a theory it had to do with my embedded environment (using `newlib`), and shouldn't affect a typical setup. I compiled a short test on my host and I do not see the warning, so I believe it shouldn't delay any releases.

FWIW I was able to stifle the warning (`pstdalign.h:36:0: warning: "_Alignas" redefined`) when including flatcc headers as such on my target, didn't have time yet to look into where to mark those `#define`s unneeded within flatcc.

```
#if __has_include(<sys/cdefs.h>)
#undef _Alignas
#undef _Alignof
#endif

extern "C" {
#include "flatcc/flatcc_builder.h"
#include "flatcc/flatcc_json_parser.h"
#include "flatcc/flatcc_verifier.h"
}
```

mikkelfj:
huh it might be that no users actually have stdalign, except for C++11. There are many C11 features missing in glibc.

mikkelfj:
Hmm your compiler must match one of the of cases in pstdalign.h, otherwise you would get
`#error please update pstdalign.h with support for current compiler`
but detecting is somewhat off. If you can think of a criteria, we can add it to pstdalign.h that is the point of that file exactly.


stdalign.h is supposed to define `__alignas_is_defined` but since it isn't you could define at as a workaround, or perhaps include stdalign.h before including flatc source.

mikkelfj:
https://lists.gnu.org/archive/html/bug-gnulib/2015-08/msg00003.html

mikkelfj:
https://lists.gnu.org/archive/html/bug-gnulib/2015-08/txtvn2XUKr9za.txt

mikkelfj:
Could you try the following before including flatcc headers?

```
#if defined(__cplusplus) && __cplusplus > 201103
#include <stdalign.h>
#define __alignas_is_defined 1
#define __alignof_is_defined 1
#endif
```

mikkelfj:
On other issue - if you have something on an esoteric system, now would be a good time to fix it. flatcc is supposed to work on many systems.

I have a new possibly breaking change I need to figure out before release anyway.

mikkelfj:
https://github.com/dvidelabs/flatcc/tree/alignas_cplusplus

mikkelfj:
https://github.com/dvidelabs/flatcc/commit/1f1661c0106e4f1a70e7b3053e71b3616f9df5b4

paulreimer:
Odd that the GitHub e-mails are so delayed, but I just tried your suggested change in my code before including flatcc headers and it worked.

But the branch above didn't work for me... I still need the workaround.

PS There is a tiny bit of trailing whitespace in the commit above. Actually I've seen some around the codebase, not a big deal but it highlights red in my editor.

paulreimer:
Perhaps because I still need to include the flatcc headers in `extern "C"`, so when I put the workaround outside the `extern "C"` it seemed to work.

mikkelfj:
I try to strip whitespace, my editor also makes a note in the status line, but it regurlary overlook them.

It's strange that the branch does not work, but since other C++ users apparently have been happy with the current form, I should probably just add those lines first in pstdalign.h

Could you try to modify the pstdalign.h to your liking?

mikkelfj:
BTW: I'm all for adding adding extern C in flatcc headers. I'm just not keen on doing it indiscriminately in portable without understanding the details.

paulreimer:
Fair enough! If there is a consistent rule to apply, I'm happy to make the patch (sounds like, "put guards *after* `#include`s, in any top-level headers in `include/flatcc/*.h`, not but within `portable/`). But if it takes understanding the context for each one then I might take awhile.

mikkelfj:
I can also do it if you test it. As I  said, I didn't want to do it without a test case. I could of course make a C++ project, but work stacks up.

mikkelfj:
BTW: why do you include flatcc headers directly. Is it due to library logic that does not know what schema will be compiled? Otherwise all the necessary `flatcc_` headers ought to be included with the generated headers.

>extern "C" {
#include "flatcc/flatcc_builder.h"
#include "flatcc/flatcc_json_parser.h"
#include "flatcc/flatcc_verifier.h"
}

paulreimer:
Yup exactly. User specifies the generated flatcc functions as rvalue references (C++ templated) when calling my library's API.

mikkelfj:
Everything merged.

mikkelfj:
Have you seen the changes I made to test_json.c that makes this modular? Sounds like you have.

https://github.com/dvidelabs/flatcc/blob/544cf78afdc6ea27330d3a981ab020972641eb72/test/json_test/test_json.c#L17-L22

I should warn you that I might be breaking union vector parser interface at some point - not sure if it is now or later. You are not likely to call it directly, but currently it has no type parser and does not yet need it.

paulreimer:
So awesome. This was WAY better than my experience using the C++ flatbuffers implementation (I needed reflection + binary schemas to accomplish the same thing there, although I have a two-pass implementation now). Now it's all in the codegen! Also, C++ flatbuffers carelessly depends on C++ iostreams, which added an unbelievable 200k to my embedded system binaries.

mikkelfj:
Thansk. But the flatcc parser can also grow large for large schema where C++ stays bounded by the implementation. I could improve on this, but it would be slower and isn't really important. You need to stay at -O2 typically, at least on Intel, or the instruction cache gets too much work.

mikkelfj:
As to two-level builds, this is one reason I'm looking into Meson build. It can quickly and reliable recompile changes in a schema included 3 levels down, or more. This in part because flatcc supports dependency generation.

paulreimer:
@mikkelfj Just to clarify but do you mean the flatcc implementation is unbounded in program size (not memory/RAM size), as the schema grows larger? But this is only for the JSON parser/printer, not using builder directly (hopefully)?

Also, what I meant about having a two-pass approach was actually in my implementation for my JSON->FB parser implementation (streaming JSON with lots of sub-objects, extracting each relevant object and parsing its JSON into a flatbuffer).

Before, when using the C++ flatbuffers library, I drove the flatbuffers reflection->builder machinery myself within the streaming JSON parser (so, single pass). Now, with flatcc, in the streaming JSON parser I buffer sub-objects into a separate JSON string and pass that to the flatcc JSON parser (which then has a complete small object to parse). So with flatcc I have this intermediate copy of a JSON string which I didn't have before. Probably using the reflection machinery in flatcc from the streaming JSON parser would be more efficient, but I am happy to avoid it for now.

mikkelfj:
>  Just to clarify but do you mean the flatcc implementation is unbounded in program size (not memory/RAM size), as the schema grows larger? But this is only for the JSON parser/printer, not using builder directly (hopefully)?

Yes that is correct. So for embedded uses with lots of SSD it is not an issue. The amount of code is also not an issue on simpler processors, but x64 with instruction caches etc. do suffer from large code sizes. But the parser runs at 300+MB/s so on a 2.2 GHz core i7 so ...
The parser is very conservative in memory use, and except for unions and a bit of recursion overhead only uses what the builder would use. Unions need some extra memory which is handled by a user stack allocated by the builder.

As to copying vs. streaming. Depending on the hardware, memcpy is much faster than parsing, so this is not an overhead considering a streaming parser would typically be slower. However, you need to have enough memory and there is some latency in starting the process.

The parser can be adapted to streaming operation with some effort, but you still need memory for building the buffer.  You can actually already today stream the output if you use non-clustered vtables and modify the emitter to flush partial content and reassemble the pages later, but it is only really good if adopted to StreamBuffers which are similar to FlatBuffers, but build front to back rather than back to front. As it is you need to revert the order of the emitter pages when the total size is known, but it keeps memory down for large documents.

mikkelfj:
If you are willing to provide a read_more_data callback, and this callback is permitted to block until data arrives, for example using a co-routine library or threads, it is fairly easy to do streaming. In a later release all the hard coded end checks will be replaced with macros that can hook into a callback. But a proper streaming parser should implement a statemachine with a stack so it can exit and resume as data arrives. But co-routines would actually be more efficient, I suppose.

mikkelfj:
I have been giving this streaming parser concept some more thoughts. I think it is possible to do efficiently and will consider implementing it in a future release. It may slow down the parser slightly though.

PR #78, opened 2018-02-24
Some more typecasts to satisfy the latest version of the clang++ compiler

AndrewJDR:
Adding these typecasts prevents the more recent versions of the clang++ compiler from complaining.


AndrewJDR:
Also see related discussion here https://github.com/dvidelabs/flatcc/pull/59

mikkelfj:
Thanks. We could use a test for running samples/monster or similar in a C++ compiler as part of the test suite to catch these early.

Issue #79, opened 2018-02-25
Add various versions of clang/clang++ to the travis CI build


AndrewJDR:
Various versions of clang/clang++ should probably be added to the travis.yml to catch issues like these:
https://github.com/dvidelabs/flatcc/pull/78
https://github.com/dvidelabs/flatcc/pull/59

I may do it at some point if i get some time, but it could be a while :/

In the meantime, I did do some research and found some good starting points:
* https://stackoverflow.com/questions/29312015/building-with-more-than-one-version-of-a-compiler/32127147#32127147
* https://github.com/taocpp/PEGTL/blob/master/.travis.yml

mikkelfj:
This should probably be done on the ci-more branch where many more gcc and clang compilers are tested.

But it won't do much without a CMake C++ test project in the `test` folder that executes when `scripts/test.sh` is being run. It could perhaps just be a C++ file that includes `samples/monster/monster.c`. Such a project should be tested against v0.5.1 to verify that it indeed triggers the mentioned warnings.

mikkelfj:
Fixed in #80 and related recent commits.

Issue #77, opened 2018-02-06
Build warning with embedded toolchain (Espressif esp-idf)


paulreimer:
```
In file included from ./lib/flatcc/include/flatcc/flatcc_flatbuffers.h:24:0,
                 from ./lib/flatcc/include/flatcc/flatcc_builder.h:68,
                 from ./src/json_to_flatbuffers_converter.h:20,
                 from ../main/rxcpp_test_task.cpp:21:
./lib/flatcc/include/flatcc/portable/pstdalign.h:36:0: warning: "_Alignas" redefined
 #define _Alignas(t) __attribute__((__aligned__(t)))
```

Similar with `_Alignof`.

In my file which includes `flatcc` headers directly (for generic JSON parsing), I just include these files without any other flags/guards:
```
#include "flatcc/flatcc_builder.h"
#include "flatcc/flatcc_json_parser.h"
#include "flatcc/flatcc_verifier.h"
```

mikkelfj:
As I recall from other issue https://github.com/dvidelabs/flatcc/pull/73#issuecomment-357337934
this is specific to C++14 because the stdalign is no longer a macro.

Could please help figure out why the proposed change to the pstdalign.h did not work for you?

https://github.com/dvidelabs/flatcc/tree/alignas_cplusplus
https://github.com/dvidelabs/flatcc/commit/1f1661c0106e4f1a70e7b3053e71b3616f9df5b4

paulreimer:
Ah right, sorry. I had tried that branch during the previous C++ work, but it didn't work at the time. Now that branch, updated with the latest `master` merged in, works fine for me!

paulreimer:
Haha wait a sec. That branch does remove the error for my C++ files which include the flatcc files.

BUT it introduces a warning now when compiling the flatcc C-files in `src/runtime/`

```
In file included from ./lib/flatcc/include/flatcc/flatcc_flatbuffers.h:24:0,
                 from ./lib/flatcc/include/flatcc/flatcc_builder.h:68,
                 from ./lib/flatcc/src/runtime/builder.c:21:
./lib/flatcc/include/flatcc/portable/pstdalign.h:48:0: warning: "_Alignas" redefined
 #define _Alignas(t) __attribute__((__aligned__(t)))
 ^
In file included from /esp-idf/components/newlib/include/stdlib.h:19:0,
                 from ./lib/flatcc/src/runtime/builder.c:16:
[...]

PR #80, opened 2018-02-25
Added a C++ test.

AndrewJDR:
This simply consists of a cpp file that includes samples/monster/monster_test.c.
This can help us detect issues like those addressed by f503a8c and cd727d0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 #cmake_minimum_required (VERSION 2.8.11)
 cmake_minimum_required (VERSION 2.8)
 
-project (FlatCC C)
+project (FlatCC C CXX)
 
 #
 # NOTE: when changing build options, clean the build using on of:
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -2,6 +2,7 @@
 # Under git such files are protected with the `.gitattributes` file.
 # Incorrect line endings may lead to failed tests.
 if (FLATCC_TEST)
+add_subdirectory(monster_test_cpp)
 add_subdirectory(cgen_test)
 add_subdirectory(monster_test)
 add_subdirectory(monster_test_solo)
--- a/test/monster_test_cpp/CMakeLists.txt
+++ b/test/monster_test_cpp/CMakeLists.txt
@@ -1,0 +1,24 @@
+include(CTest)
+
+# Note: This re-uses the samples/monster fbs and .c file.
+
+set(INC_DIR "${PROJECT_SOURCE_DIR}/include")
+# We use our own separate gen dir so we don't clash with the real monster sample.
+set(GEN_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated/monster_test_cpp")
+set(FBS_DIR "${PROJECT_SOURCE_DIR}/samples/monster")
+
+include_directories("${GEN_DIR}" "${INC_DIR}")
+
+add_custom_target(gen_monster_test_cpp ALL)
+add_custom_command (
+    TARGET gen_monster_test_cpp
+    COMMAND cmake -E make_directory "${GEN_DIR}"
+    COMMAND flatcc_cli -a -o "${GEN_DIR}" "${FBS_DIR}/monster.fbs"
+    DEPENDS flatcc_cli "${FBS_DIR}/monster.fbs"
+)
+
+add_executable(monster_test_cpp monster_test.cpp)
+add_dependencies(monster_test_cpp gen_monster_test_cpp)
+target_link_libraries(monster_test_cpp flatccrt)
+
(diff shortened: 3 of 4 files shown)


AndrewJDR:
Hmm, I'm not entirely sure what's going wrong on travis.

If i had to guess, it's using a the c++ compiler even on c files. If so, I'm not sure how to prevent that from happening using cmake.

Also note:
I found that in order to get cmake to use g++/clang++ to build the monster_test.cpp, I had to change `project (FlatCC C)` to `project (FlatCC C CXX)` in the CMakeLists.txt file in the flatcc repo root directory.

AndrewJDR:
Ok, things seem to be in order now.

The monster_test_cpp is now failing as expected due to the c++ compiler legitimately complaining about the same kinds of things addressed in f503a8c and cd727d0. I don't think I'll have time to fix all of these complaints just yet, but at least we have a test for them now.

mikkelfj:
I think it would be fine to add the test to master branch, if it is not intrusive. It is only if the extra C++ versions are needed that the ci-more branch is needed. But lets just get it working first. Note the ci-more is regularly behind because it is mostly updated prior to a significant release.

mikkelfj:
We could perhaps have a FLATCC_TEST_CPP option in the main CMake file. It would probably be on by default, but could be used to switch off the test for some versions.

mikkelfj:
@paulreimer you might be interested in this as well. See also https://github.com/dvidelabs/flatcc/issues/79

AndrewJDR:
With clang 5 (Listed simply as "Clang C" on the travis page), there's some alignas complaints that don't appear in the clang 3.x travis runs. I did some poking around online, but I'm not sure exactly what to do about those yet.

mikkelfj:
Note that MSVC 2010 fails C++ due to missing stdaligh.h. This is probably because pstdalign.h does not correctly compiler detection when using C++.
https://ci.appveyor.com/project/dvidelabs/flatcc/build/win%23694/job/aqwnyjg62g9k3t8i

mikkelfj:
Our comments crossed, but it all relates https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pstdalign.h
@paulreimer did some on research on getting this working on C++14 with newlib, but there is a way to go yet for all compilers, it appears.

mikkelfj:
The MSVC 2010 is fairly obvious: pstdalign assumes that stdalign.h is available if the compiler is C++, which is a bit naive, considering the players in that market.

https://github.com/dvidelabs/flatcc/blob/de2da1ffca39d191aec160d91191ed2176a93b57/include/flatcc/portable/pstdalign.h#L34-L39

mikkelfj:
See also discussion here: https://github.com/dvidelabs/flatcc/issues/77

mikkelfj:
Another thing - samples/monster does not compile in MSVC 2010 for C because it uses C++ inline variable declarations. It might actually work when included in a C++ project, but otherwise it might be better to include https://github.com/dvidelabs/flatcc/blob/master/test/monster_test/monster_test.c

mikkelfj:
I committed the MSVC related issue - but it cannot be tested before C++ tests are in CI.

I also reproduced the alignas problem on clang MacOS.

The issue is that the built-in C++ alignas macro is broken - alignas has several forms including `alignas(type)` and `alignas(pack)`. It apparently expects a type and does not allow a numeric pack argument `alignas(4) float;` as needed by flatcc output. This can be fixed by replacing the entire pstdalign.h file with

```
include <stdalign.h>
#undef alignas
#define alignas(t) __attribute__((__aligned__(t)))
```

but if that is done, a very precise knowledge of the compiler version is required, and it might break some user-level C++ code.

Another fix is to force clang to use std=c++11 just like it is doing for C in the main make file:

```
if (CMAKE_C_COMPILER_ID MATCHES "Clang")
    # Clang or AppleClang
    message(STATUS "Setting Clang compiler options")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -pedantic -Wall -Wextra")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pedantic -Wall -Wextra")
...
```

For now, I'm inclined to just clamp C++ to std=c++11.

With these fixes, the monster test fails with the well-known null pointer assignments for unions, as expected.

mikkelfj:
I added PORTABLE_PATCH_CPLUSPLUS_STDALIGN flag in CMakeLists.txt main file, and configured C++ to use std=c++11. Either of these should work, but enabled c++11 version.

@paulreimer you might want to recheck newlib, since stdalign has been messed with.

Several warnings fixed for clang/MSVC in C++ mode.

https://ci.appveyor.com/project/dvidelabs/flatcc/build/win%23704
https://travis-ci.org/dvidelabs/flatcc/builds/351946954

PR #86, opened 2018-06-06
Fix build with gcc-8

baruchsiach:
gcc 8 enables a warning about strncpy that might leave the destination
string without a terminating null character. This breaks the build:

In function ‘read_thash_identifier’,
    inlined from ‘flatcc_verify_buffer_header’ at .../src/runtime/verifier.c:468:15:
.../src/runtime/verifier.c:114:5: error: ‘strncpy’ specified bound 4 equals destination size [-Werror=stringop-truncation]
     strncpy((char *)&id, identifier, sizeof(id));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

.../src/runtime/json_printer.c: In function ‘accept_header’:
.../src/runtime/json_printer.c:1023:9: error: ‘strncpy’ specified bound 4 equals destination size [-Werror=stringop-truncation]
         strncpy((char *)&id2, fid, FLATBUFFERS_IDENTIFIER_SIZE);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

.../src/compiler/codegen_schema.c: In function ‘flatbuffers_has_identifier.constprop’:
.../include/flatcc/reflection/flatbuffers_common_reader.h:436:3: error: ‘strncpy’ output truncated before terminating nul copying 4 bytes from a string of the same length [-Werror=stringop-truncation]
   strncpy((char *)&id2, fid, sizeof(id2));
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Use memcpy instead, since memcpy is not expected to leave a null
terminator.

### DIFF
--- a/include/flatcc/reflection/flatbuffers_common_reader.h
+++ b/include/flatcc/reflection/flatbuffers_common_reader.h
@@ -433,7 +433,7 @@ static inline T N ## _ ## NK(N ## _struct_t t) { return t ? &(t->NK) : 0; }
 /* If fid is null, the function returns true without testing as buffer is not expected to have any id. */
 static inline int flatbuffers_has_identifier(const void *buffer, const char *fid)
 { flatbuffers_thash_t id, id2 = 0; if (fid == 0) { return 1; };
-  strncpy((char *)&id2, fid, sizeof(id2));
+  memcpy(&id2, fid, sizeof(id2));
   /* Identifier strings are always considered little endian. */
   id2 = __flatbuffers_thash_cast_from_le(id2);
   id = __flatbuffers_thash_read_from_pe(((flatbuffers_uoffset_t *)buffer) + 1);
--- a/src/runtime/json_printer.c
+++ b/src/runtime/json_printer.c
@@ -1020,7 +1020,7 @@ static int accept_header(flatcc_json_printer_t * ctx,
         return 0;
     }
     if (fid != 0) {
-        strncpy((char *)&id2, fid, FLATBUFFERS_IDENTIFIER_SIZE);
+        memcpy(&id2, fid, FLATBUFFERS_IDENTIFIER_SIZE);
         id2 = __flatbuffers_thash_cast_from_le(id2);
         id = __flatbuffers_thash_read_from_pe((uint8_t *)buf + offset_size);
         if (!(id2 == 0 || id == id2)) {
--- a/src/runtime/verifier.c
+++ b/src/runtime/verifier.c
@@ -111,7 +111,7 @@ static inline uoffset_t read_uoffset(const void *p, uoffset_t base)
 static inline thash_t read_thash_identifier(const char *identifier)
 {
     flatbuffers_thash_t id = 0;
-    strncpy((char *)&id, identifier, sizeof(id));
+    memcpy(&id, identifier, sizeof(id));
     return __flatbuffers_thash_cast_from_le(id);
 }
 


mikkelfj:
Thanks, but this won't work if the source string is less than 4 characters.

```
const char *s;
int i;

for (s = fid, i = 0; i < sizeof(id2); ++i) {
    id2[i] = *s;
    s += *s != 0;
}
```
This both handles fid less than 4 bytes without buffer overrun and ensure that trailing data is zeroed which the currently solution might not do?

Note that variables must always be declared first in a function otherwise MSVC 2010 gets sad.

mikkelfj:
The proper solution is probably to use

flatbuffers_type_hash_from_identifier
in the flatcc_identifier.h
and add
flatbuffers_type_hash_from_string based on the above suggested function.

Then use either of these functions elsewhere depending on the input type.

arnout:
Instead of these hacks, maybe just locally disable the warning? Cfr. https://stackoverflow.com/questions/50198319/gcc-8-wstringop-truncation-what-is-the-good-practice

    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wstringop-truncation"
        strncpy(d, s, 32);
    #pragma GCC diagnostic pop

mikkelfj:
Yes, I was considering this as well. But generally I prefer to keep that warning list short if there is a clean workaround. In this case it is logical to have a function that converts a string id to an integer id. Warnings also affect application code. That said, I'm getting pretty tired of things breaking fundamentally in every new gcc release when the affected code is perfectly valid and common C.

mikkelfj:
For reference, warnings and diagnostics are handled here

https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_prologue.h
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pdiagnostic_push.h
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pwarnings.h

and in the build system if that is not practical.

include/flatcc/reflection/flatbuffers_common_reader.h
```diff
@@ -433,7 +433,7 @@ static inline T N ## _ ## NK(N ## _struct_t t) { return t ? &(t->NK) : 0; }
 /* If fid is null, the function returns true without testing as buffer is not expected to have any id. */
 static inline int flatbuffers_has_identifier(const void *buffer, const char *fid)
 { flatbuffers_thash_t id, id2 = 0; if (fid == 0) { return 1; };
-  strncpy((char *)&id2, fid, sizeof(id2));
+  memcpy(&id2, fid, sizeof(id2));
```
mikkelfj: This is generated code, so the fix should be placed in the code generator.

mikkelfj:
@baruchsiach can you please confirm that c0df0b6fca4fa6bca114bba4df74f1b4e53124c0 fixes the issue. I do not have gcc-8 readily available.

mikkelfj:
I see this is blocking a build root packet. I can make a new release soon if need be, probably should anwyay.

baruchsiach:
@mikkelfj I verified that commit c0df0b6fca4fa6bca114bba4df74f1b4e53124c0 fixes the build failure with gcc 8, and posted a patch to the Buildroot mailing list.

mikkelfj:
Thanks for the update.

mikkelfj:
There is a new flatcc 0.5.2 release - I have not testet with gcc 8 but the above commit should be included.

PR #91, opened 2018-08-09
Fix switch-case statement

jesferman:
Hi,

Correct me if I am missing something, but I think these two `break` are necessary.


mikkelfj:
Thanks, yes it indeed looks like they were missing. Fortunately an error would have been generated if that code path were executed and even more furtunately, it appears that the function is never called.

PR #94, opened 2018-09-06
CMakeLists.txt: conditionally require C++ based on FLATCC_TEST

JoelsonCarl:
C++ is only used when building the tests, so only include it as a
language via project() when FLATCC_TEST is enabled. This allows
toolchains that don't have C++ to build flatcc.

Signed-off-by: Joel Carlson <JoelsonCarl@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,18 @@
 #cmake_minimum_required (VERSION 2.8.11)
 cmake_minimum_required (VERSION 2.8)
 
-project (FlatCC C CXX)
+# Disable build of tests and samples. Due to custom build step
+# dependency on flatcc tool, some custom build configurations may
+# experience issues, and this option can then help.
+option(FLATCC_TEST "enable tests" ON)
+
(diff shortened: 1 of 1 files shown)


JoelsonCarl:
This was found by the buildroot autobuilders, when the next branch was updated to pull in version 0.5.2 of flatcc.  One of the builds used a toolchain that did not include C++ support.  So I see there being two options:

1. patch things over here so that you can support toolchains without C++ (by not requiring it if not building the tests)
2. If you don't like this at all, I'll have to patch the buildroot configs to only allow flatcc to be included if the toolchain has C++.

mikkelfj:
Thanks

mikkelfj:
For reference, C++ test was added because there are several C++ users and compiler warnings differ breaking C++ builds with aggressive warnings. But it is the intention that flatcc should be entirely independent of C++. Another case where C++ is being used is in the benchmark where performance is compared to an old version of C++ flatbuffers, but this only builds with dedicated shell scripts.

PR #101, opened 2018-12-31
Added a failing test inside monster_test for a crash in _pick.

AndrewJDR:
To repro, compile and run monster_test_d.

See backtrace.txt and search for XXX inside of the following files
for important notes regarding the issue:
test/monster_test/monster_test.c
test/monster_test/pick_test.fbs

Perhaps this is an incorrect library usage somewhere, but it certainly seems odd.

I've commented out all other tests for now, mainly to simplify my workflow while working on this issue, but also because the rest of the monster tests weren't playing nice with this one when I tried to compile everything together.

### DIFF
--- a/backtrace.txt
+++ b/backtrace.txt
@@ -1,0 +1,17 @@
+Program received signal SIGSEGV, Segmentation fault.
+
+Stack
+[0] from 0x00007ffff7a91f78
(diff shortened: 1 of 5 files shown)


mikkelfj:
Closing this after using it to verify fix to #102.

PR #104, opened 2019-02-12
Added a failing test for the int key sort issue.

AndrewJDR:
This PR adds a failing test for what might be an issue with sorting on integer keys.
This uses a uint64 type as the key in the schema, but I've found that it also fails on uint32/uint16/uint8 as well.

To repro, run `scripts/test.sh  --debug`. This should be printed:
```
          3 - monster_test (SEGFAULT)
```

The backtrace looks like this:
```
#0  0x000000000040395f in __flatbuffers_voffset_read_from_pe (p=0xffffffff93f74065) at ../../include/flatcc/flatcc_endian.h:91
#1  0x000000000040ab53 in MyGame_Example_Monster_intkey_get (t__tmp=0x63b5ac) at test/monster_test/generated/monster_test_reader.h:810
#2  0x000000000040ae59 in __MyGame_Example_Monster_sort_by_intkey__heap_sift_down (vec__tmp=0x63b58c, start__tmp=0, end__tmp=2) at test/monster_test/generated/monster_test_reader.h:814
#3  0x000000000040b009 in __MyGame_Example_Monster_sort_by_intkey__heap_sort (vec__tmp=0x63b58c) at test/monster_test/generated/monster_test_reader.h:814
#4  0x000000000040b02a in MyGame_Example_Monster_vec_sort_by_intkey (vec__tmp=0x63b58c) at test/monster_test/generated/monster_test_reader.h:814
#5  0x000000000041a4c8 in test_sort_find_by_intfield (B=0x7fffffffd450) at ../../test/monster_test/monster_test.c:1211
#6  0x000000000041e61c in main (argc=1, argv=0x7fffffffd6c8) at ../../test/monster_test/monster_test.c:2860
```

### DIFF
--- a/src/compiler/codegen_c_sort.c
+++ b/src/compiler/codegen_c_sort.c
@@ -77,8 +77,8 @@ int gen_sort(fb_output_t *out)
         "#define __%sstring_diff(x, y) __%sstring_n_cmp((x), (const char *)(y), %sstring_len(y))\n",
         out->nsc, out->nsc, out->nsc, out->nsc);
     fprintf(out->fp,
-        "#define __%sscalar_swap(vec, a, b, TE) { TE x__tmp = vec[b]; vec[b] = vec[a]; vec[a] = x__tmp; }\n"
-        "#define __%sstring_swap(vec, a, b, TE)\\\n"
+        "#define __%svalue_swap(vec, a, b, TE) { TE x__tmp = vec[b]; vec[b] = vec[a]; vec[a] = x__tmp; }\n"
+        "#define __%suoffset_swap(vec, a, b, TE)\\\n"
         "{ TE ta__tmp, tb__tmp, d__tmp;\\\n"
         "  d__tmp = (TE)((a - b) * sizeof(vec[0]));\\\n"
         "  ta__tmp =  __flatbuffers_uoffset_read_from_pe(vec + b) - d__tmp;\\\n"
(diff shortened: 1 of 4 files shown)


AndrewJDR:
Possibly related, or possibly unrelated, a particular part of this PR also reproduces some json test failures.

Specifically, adding the following to the end of the monster_test.fbs:
```
intkey:uint64 (id:36, key);
```
Will result in:
```
         11 - test_json (Failed)
         12 - json_test_uql (Failed)
         13 - json_test_uql_off (Failed)
         14 - json_test_uq (Failed)
         15 - json_test_uq_off (Failed)
         16 - json_test (Failed)
```
(No need for any of the sourcecode changes to repro this... just this one-line schema addition is sufficient)

mikkelfj:
Some JSON tests likely print default values of FlatBuffers while others only print forced values. Adding an extra int field would introduce an extra 0 in the buffer and fail when compared to expected output. I don't think the failing JSON tests are related or that they are of concern.

mikkelfj:
The problem is most likely related to the swap operation. Integer vectors use a scalar swap operation which is fine for sorting integers. But table vectors with integer keys should a uoffset swap operation that is similar to string swaps. To complicate matters, vectors of structs can also be sorted on integer keys, but these should swap the entire struct (C++ FB likely doesn't support that).

uoffset_t swaps are tricky because their value depend on their location, so when stored in a new place, a difference must be subtracted. Strings do this. It is possible that sorting by string keys accidentially work because the table vector swap and the string vector swap is the same operation.

mikkelfj:
Could you please take a look at the commit I just made to your branch?

src/compiler/codegen_c_sort.c
```diff
@@ -87,12 +87,28 @@ int gen_sort(fb_output_t *out)
...
         out->nsc, out->nsc, out->nsc, out->nsc);
+    fprintf(out->fp,
+        "#define __%sdefine_table_sort_by_scalar_field(N, NK, TK)\\\n"
+        "  __%sdefine_sort_by_field(N, NK, TK, %suoffset_t, __%sscalar_diff, __%stable_swap)\n",
+        out->nsc, out->nsc, out->nsc, out->nsc, out->nsc);
     fprintf(out->fp,
         "#define __%sdefine_sort_by_string_field(N, NK)\\\n"
-        "  __%sdefine_sort_by_field(N, NK, %sstring_t, %suoffset_t, __%sstring_diff, __%sstring_swap)\n",
+        "  __%sdefine_sort_by_field(N, NK, %sstring_t, %suoffset_t, __%sstring_diff, __%suoffset_swap)\n",
         out->nsc, out->nsc, out->nsc, out->nsc, out->nsc, out->nsc);
```
mikkelfj: this should have been __%stable_swap, but it amounts to the same thing.

AndrewJDR: Passes the test as I'm sure you already know. I've also been trying it in my project and it's working fine so far there as well. :+1:

mikkelfj:
Now recommitted on master without your debug changes - please confirm there as well.

AndrewJDR:
Yes, looking good on master as well. Thanks again.

PR #109, opened 2019-03-05
Correct FLATCC_INSTALL_LIB bug

nebiun:
Replace lib with ${lib_dir} in compiler and runtime CMakeLists.

### DIFF
--- a/src/compiler/CMakeLists.txt
+++ b/src/compiler/CMakeLists.txt
@@ -39,5 +39,5 @@ endif(FLATCC_REFLECTION)
 add_library(flatcc ${SOURCES})
 
 if (FLATCC_INSTALL)
-    install(TARGETS flatcc DESTINATION lib)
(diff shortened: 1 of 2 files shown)


mikkelfj:
Shouldn't this be `${dist_dir}/${lib_dir}` - see root CMakeLists

nebiun:
No

mikkelfj:
why?

nebiun:
Original code use only 'lib' to define the destination directory, then full path is not required.
I replaced 'lib' with the variable that can contain 'lib' or 'lib64'.
No more, no less.

mikkelfj:
I'm not sure the original code was correct - what happens if $dist_dir changes?

nebiun:
dist_dir is set to ${PROJECT_SOURCE_DIR} and does not affect DESTDIR that is the "root" path for the installation.

mikkelfj:
I see, so dist_dir is a bit misleading - probably from before install support was added to the project.

PR #114, opened 2019-03-29
Reorderd member mistake when 1st field is max_id

charleszhangjie:
write a schema to define a table which the first field has the max_id, flatcc falls into endless loop again.
e.g.

table x {
     b:int (id:2);
     a:int (id:0);
     c:int (id:1);
   } 
the original member list is 2->0->1 , the reordered member is 0->1->2 ( field with id:2 must link to null instead of id:0) unless a loop occurs in the list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
 - Use `FLATCC_INSTALL_LIB` for install targets in addition to ordinary builds
   (#109).
 - Add missing case break in flatcc compiler - no impact on behavior (#110).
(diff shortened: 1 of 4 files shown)


mikkelfj:
This goes wrong when max_id is 0. I already thought about that, but it ought to be protected by `count > 0` earlier.

bin/flatcc_d -av --json -o build/Debug/test/json_test/generated test/monster_test/monster_test.fbs

mikkelfj:
Change included in separate commit 5fb4666 that also deals with build error.
The above does not work by itself directly because max_id is not valid when need_id is false.

PR #115, opened 2019-04-15
Fix commandline in README

keroro520:
`flatcc -cr` only generates common_reader, but not the protocol reader. So I think `flatcc -cr --reader` to also generate protocol reader is better for introducing "Reading a Buffer".

### DIFF
--- a/README.md
+++ b/README.md
@@ -775,14 +775,14 @@ it is an adapted extract of the [monster_test.c] file.
 First we compile the schema read-only with common (-c) support header and we
 add the recursion because [monster_test.fbs] includes other files.
 
-    flatcc -cr test/monster_test/monster_test.fbs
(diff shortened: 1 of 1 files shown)


mikkelfj:
I wonder if it is a bug in the command line parser.

keroro520:
> I wonder if it is a bug in the command line parser.

I don't think it is an implementation bug.
In help document [here](https://github.com/dvidelabs/flatcc/blob/master/src/cli/flatcc_cli.c#L74):

> --reader is the default option to generate reader output but can be used
>        explicitly together with other options that would otherwise disable it

Then if given option `-cr`, parser would treat it as shortcut of `--common --receusive`.

mikkelfj:
Yes you right, thanks!

PR #120, opened 2019-07-10
Fix misc build errors on GCC 9

tonyhutter:
Fix the following build errors on Fedora 30:
```
src/compiler/semantics.c:702:20: error: this statement may fall through
  702 |             key_ok = 0;

samples/reflection/bfbs2json.c:86:5: error: ‘%s’ directive argument is null

test/monster_test/monster_test.c:1948:5: error: ‘strncpy’ output truncated copying 5 bytes from a string of length 12
 1948 |     strncpy(foobar->text, "hello, world", ns(FooBar_text_get_len()));
```

I've done zero testing of this other than it builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,20 @@ elseif (CMAKE_C_COMPILER_ID STREQUAL "GNU")
     else()
         message(STATUS "Setting GNU C compiler options with c11 and Posix")
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -pedantic -Wall -Wextra")
+        if (GCC_VERSION VERSION_GREATER_EQUAL 8.0)
+            # Disable some GCC checks:
+            #
+            # -Wstringop-truncation:
+            # strncpy into char arrays in FlatBuffer structs, these are
+            # zero-paddded but not zero terminated
+            #
+            # -Wno-format-overflow:
+            # GCC 9 warns on mistakenly assumed NULL string when 
+            # printing from a required FlatBuffer string field.
+            #
+            message(STATUS "Disabling GNU C compiler warnings: -Wstringop-truncation -Wno-format-overflow")
+            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-stringop-truncation -Wno-format-overflow")
+        endif()
         if (FLATCC_GNU_POSIX_MEMALIGN)
             # -std=c11 prevents detection of posix_memalign and aligned_alloc might be missing
             set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DPORTABLE_POSIX_MEMALIGN=1")
--- a/src/compiler/semantics.c
+++ b/src/compiler/semantics.c
@@ -523,7 +523,7 @@ static int analyze_struct(fb_parser_t *P, fb_compound_type_t *ct)
         member = (fb_member_t *)sym;
         switch (member->type.type) {
         case vt_fixed_array_type:
-            /* -fallthrough */
+            /* fall through */
         case vt_scalar_type:
             t = member->type.t;
             member->type.st = map_scalar_token_type(t);
@@ -536,7 +536,7 @@ static int analyze_struct(fb_parser_t *P, fb_compound_type_t *ct)
             member->size = size * member->type.len;
             break;
         case vt_fixed_array_compound_type_ref:
-            /* -fallthrough */
+            /* fall through */
         case vt_compound_type_ref:
             /* Enums might not be valid, but then it would be detected earlier. */
             if (member->type.ct->symbol.kind == fb_is_enum) {
@@ -700,7 +700,7 @@ static int process_struct(fb_parser_t *P, fb_compound_type_t *ct)
         switch (member->type.type) {
         case vt_fixed_array_type_ref:
             key_ok = 0;
-            /* -fallthrough */
+            /* fall through */
         case vt_type_ref:
             type_sym = lookup_type_reference(P, ct->scope, member->type.ref);
             if (!type_sym) {


mikkelfj:
I think one of the errors is that GCC 9 does not understand "/* -fallthrough */", but probably does understands "/* fall through */", could you please test that?

tonyhutter:
Yep, `/* fall through */` works.  I included that change in my latest push.

test/monster_test/monster_test.c
```diff
@@ -1945,8 +1945,9 @@ int test_fixed_size_array(flatcc_builder_t *B)
     foobar->tests[0].b = 4;
     foobar->tests[1].a = 1;
     foobar->tests[1].b = 2;
-    strncpy(foobar->text, "hello, world", ns(FooBar_text_get_len()));
-    // or strncopy(foobar->text, "hello, world", sizeof(foobar->text));
+    strncpy(foobar->text, "hello, world", sizeof(foobar->text));
+    foobar->text[sizeof(foobar->text) - 1] = '\0';
+
```
mikkelfj: We are testing that we are not writing too much and that we use all of the 5 characters available. Hence the zero termination should not be there. I'm not sure how to silence that warning. There are many valid use cases for that syntax. We are also testing that the get_len() function is present and does it's job, so removing it does  no good. Maybe we need to truncate the string to "hello". Why does GCC have to be this way, every year.

samples/reflection/bfbs2json.c
```diff
@@ -81,9 +81,15 @@ void print_object(reflection_Object_table_t O)
     reflection_Field_vec_t Flds;
     reflection_Field_table_t F;
     size_t i;
+    const char *name;
 
     Flds = reflection_Object_fields(O);
-    printf("{\"name\":\"%s\"", reflection_Object_name(O));
+
```
mikkelfj: This is why IBM added a slash or a dot to their zeroes. The argument is the letter following P in uppercase, not a zero, or it should have been. but I can't really tell. This is just a stupid argument. What happens if you change O to Obj?

Certainly your fix is safer, but unnecessary because a valid input buffer cannot have a null name, and if the buffer is not valid, there are 700 other places to crash harder than here.

tonyhutter: > What happens if you change O to Obj?
```
flatcc/samples/reflection/bfbs2json.c:86:55: error: ‘Obj’ undeclared (first use in this function)
   86 |     printf("{\"name\":\"%s\"", reflection_Object_name(Obj));
```

mikkelfj: You need to change the argument name of the function as well, and the other places where it is being used in the function.

mikkelfj:
I'm getting the following from an search for `-fallthrough`:

flatcc$ ag fallthrough
CHANGELOG.md
155:- Silence gcc 7.x warnings about implicit fallthrough (#61).

src/compiler/semantics.c
526:            /* -fallthrough */
539:            /* -fallthrough */
703:            /* -fallthrough */

I actually made the change from `fall through` to `-fallthrough` because clang doc said is was some more correct than ordinary comment that was there before, and __attibute__ isn't portable. But that was not to be.

test/monster_test/monster_test.c
tonyhutter: I tried changing this to "hello", but:
```
error: ‘strncpy’ output truncated before terminating nul copying 5 bytes from a string of the same length [-Werror=stringop-truncation]
 1948 |     strncpy(foobar->text, "hello", sizeof(foobar->text));
```
I'm not really sure what this is testing, so I'm open to any ideas on how to fix it.

tonyhutter:
Latest push:
- renamed arg `O` to `Obj`
- renamed all `-fallthrough` to `fall though`
- changed `"hello, world"` to `"test"`

mikkelfj:
Does thiis pass on GCC 9?

I'm still not happy with this - I see why hello seems truncated, but that is what I want to test. GCC 9 should not decide whether I zero terminate my buffers in a data format with size constraints. Your original fix is probably also only a temporary solution because GCC did not get around to make that check yet. I'm inlclined to reduce the warning level of GCC.

samples/reflection/bfbs2json.c
```diff
@@ -76,14 +76,20 @@ void print_attributes(reflection_KeyValue_vec_t KV)
...
-    Flds = reflection_Object_fields(O);
-    printf("{\"name\":\"%s\"", reflection_Object_name(O));
+    Flds = reflection_Object_fields(Obj);
+
+    name = reflection_Object_name(Obj);
+    if (!name)
+        name = "NULL";
+
+    printf("{\"name\":\"%s\"", name);
     printf(",\"fields\":[");
```
mikkelfj: This check should be reverted back to the original form - both because the change is, or should be, unnecessary, and because we need to understand what is going on if there really is a problem and not just a mistyped letter.

test/monster_test/monster_test.c
mikkelfj: The test is partially a documention of a feature, partially a test.
As documentation it shows how strncpy fills an array with characters padded with zeroes up to the specified length. Char arrays in FlatBuffers are not required to be zero terminated, but they are required to be zero padded. strncpy implements these semantics perfectly.

In the concrete test we test the if the input is longer (truncation), we get the expected prefix without zero termination. A later part of the test reads back the buffer and checks for the 5 bytes "hello". That is, we can store a string without zero padding.

(Whether users want to use zero termination or not, is up to the user, but in this case it is not dictated by the format because we want the array structure to be able to hold arbitrary data and to be compatible with the uint8 array).

The write part of the test goes:
1. Given a a char array of length 5. 
2. Test that the get_len() function is present and can be used as arg to strncopy.
    (We do not test get_len() == 5, but we could).
3. We test the strncopy copies exactly get_len() bytes into the array. (we can't do this perfectly as the buffer might hold padding, but at least it must not overwrite valid data in the buffer and it must write all data expected.

We document in a comment that sizeof can also be used because the array is indeed implemented as a C-array, but this is not what we are testing.

I don't know how to fix this with GCC 9 because I don't have GCC 9, but I do want this to work with strncpy in general, because it does the right thing for this use case.

I am considering reducing the warning level of GCC because I spend so much time on counterproductive warnings. Yet GCC does not find 32/64 mixed types that Windows MSVC is complaining about as a warning (and which I should fix).

mikkelfj: Depending on how familiar you are with Flatbuffers: strings are normally stored as separate objects with their own length prefix field and a trailing zero terminator always. Char arrays is new, not yet released feature along with other fixed length arrays in structs. The Google flatbuffers project is still debating whether there should be a char array, or it should be an uint8 with an attribute to mark it for use with string content. Either way, flatcc would end up with the same char array and be compatible with uint8 arrays. Other languages might not have a char array concept and would map this to uint8 arrays regardless - also a reason to not force zero termination.

mikkelfj:
Could you please test if you can disable `-pedantic`, or some other change in flags, for GCC >= 9 

Here is the current logic in CMake which makes exceptions for older GCC.

https://github.com/dvidelabs/flatcc/blob/d42940d864ac2d1d6d90fba8776382062b22e185/CMakeLists.txt#L188-L203

derosier:
On Thu, Jul 11, 2019 at 1:10 AM MikkelFJ <notifications@github.com> wrote:

> Could you please test if you can disable -pedantic, or some other change
> in flags, for GCC >= 9
>
> Here is the current logic in CMake which makes exceptions for older GCC.
>
>
> https://github.com/dvidelabs/flatcc/blob/d42940d864ac2d1d6d90fba8776382062b22e185/CMakeLists.txt#L188-L203
>

Newer versions of GCC implement a number of new warnings - all with the
goal of improving code. Basically, it's moving some static analysis into
the compiler. In particular the Linux kernel has been finding bugs and
implementing a ton of fixes because of these additions.  The reason they're
warnings is because technically it's legal code, it just "smells" wrong and
is likely an error. It might be valid code, as it is in this case.

The typical pattern would to be to keep the warnings on but mark the single
case to disable the warning for that piece of code.  It basically gives you
the safety-net with silencing the warning with an annotation of "yes, I
intend to do this here on purpose".  This is done with a pragma in the form
of:

    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wno-format-contains-nul"
      foo(b);                       /* no diagnostic for this one */
    #pragma GCC diagnostic pop

https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html
https://gcc.gnu.org/onlinedocs/gcc-4.6.0/gcc/Warning-Options.html

Note, for the above example, I just picked a likely warning as I couldn't
find the actual gcc warning output in this thread. You'll need to pick the
right one for what you're trying to do.

@ the original poster - "compiles OK but I didn't test it" is not
acceptable. Your suggested patches would have broken the code for everyone.
This project has a test suite - use it.

- Steve

mikkelfj:
>  #pragma GCC diagnostic ignored "-Wno-format-contains-nul"

This is what I mean by other flags. If you have a concrete suggestion I'm all ears.

I do appreciate static analysis and use it a lot with clang. But I'm not convinced that GCC is striking the right balance. If you could argue that you _only_ wrote code for GCC that is one thing - then you can sprinkle __attribute__ all over the place, but it is really hard to write portable code that includes GCC these days. The consequence may be that warnings just gets disabled, that appears to be common these days.

mikkelfj:
As you can see here https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=flatcc
AUR disabled warnings for flatcc, not because the build was broken per se, but because GCC is a moving target, and they didn't want to repeatedly deal with that.

tonyhutter:
So my latest push just changes it to a `memcpy(foobar->text, "hello", sizeof(foobar->text))`.   It builds under GCC 9 and passes the test suite.  Will that work?

mikkelfj:
Well, it does the job, but fails to illustrate the use of strncpy, which is apparently no longer part of the standard library. I'd much prefer if we could silence the GCC warning somehow.

derosier:
On Thu, Jul 11, 2019 at 9:23 AM MikkelFJ <notifications@github.com> wrote:
>
> #pragma GCC diagnostic ignored "-Wno-format-contains-nul"
>
> This is what I mean by other flags. If you have a concrete suggestion I'm all ears.
>

Well, I'm not building it and getting those errors, so I don't know
which flags are the issue here.  If the specific warning text from GCC
was posted here, I'd be happy to give it a go.


> I do appreciate static analysis and use it a lot with clang. But I'm not convinced that GCC is striking the right balance. If you could argue that you only wrote code for GCC that is one thing - then you can sprinkle attribute all over the place, but it is really hard to write portable code that includes GCC these days. The consequence may be that warnings just gets disabled, that appears to be common these days.
>

Trying to target multiple toolchains perfectly is a thankless task.
And, agreed, GCC is a moving target.

My use-case is different - I support my environment and tools and the
toolchain doesn't change unless I choose to do so. And typically, we
don't upgrade toolchains in embedded work as that can be dangerous. In
fact, we even archive the specific version connected to each
production build.  Which is all a different environment than an open
source library that is used by anyone on any toolchain.

My suggestion - choose a specific toolchain version you want to
support. Keep enabled the warnings that are useful. Globally disable
ones that frequently get triggered by the type code that you must
write, as we know the library has to do a few things that are too
tricky for GCC to like. And then in the README (or somewhere) say,
"Hey, we use GCC vX.Y.Z as our standard toolchain. It works and
compiles cleanly with no warnings. Other versions of GCC (or clang, or
msvc++ or whatever) may or may not throw warnings due to changes in
the toolchain".  And then call it good. And only accept "the compiler
does it wrong" patches if it's valid for the published GCC version.
The nice thing about GCC is it's trivial to get a specific version and
stick with it, even if the distro's version changes.

Another way to handle it - have a couple of build targets that have
the full warnings enabled so you can take advantage of it when you
want do development, but have the default build targets that the users
of the library use that turn them off.  That way you get the use of
them when working with it, but end-users aren't bothered by
tool-chain-version-dependant spurious warnings.

Again - I'm not trying to develop a library that can be used by random
people on random toolchains. So, no criticism was intended. I'm just
spitballing ideas to try to help.

- Steve

tonyhutter:
Discussions about GCC aside, would you guys be ok with @derosier suggestion of simply disabling the warning around the strncpy() for the time being? 
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wno-format-contains-nul" foo(b); /* no diagnostic for this one */
#pragma GCC diagnostic pop
```
I'm just trying to reach a consensus on what I should do.

mikkelfj:
>Well, I'm not building it and getting those errors, so I don't know
which flags are the issue here.  If the specific warning text from GCC
was posted here, I'd be happy to give it a go.

I'd appreciate if you can reproduce and possible find the proper flag til kill or set, `-pendantic` or otherwise.

The posting at the top of this issue mentions the following error for GCC 9:

```
test/monster_test/monster_test.c:1948:5: error: ‘strncpy’ output truncated copying 5 bytes from a string of length 12
 1948 |     strncpy(foobar->text, "hello, world", ns(FooBar_text_get_len()));
```

Note that the test specifically is testing truncation with strncpy: you have a long string, you want til fill the arrays as much as possible, and no more.

> My suggestion - choose a specific toolchain version you want to
support.

That is not really an option since a key selling point is that it runs everywhere, and it practically does. But I spend a disproportionate amount of time on GCC and all the critical errors I find, are not from GCC warnings, but from MSVC or clang.

> Another way to handle it - have a couple of build targets that have
the full warnings enabled so you can take advantage of it when you
want do development, but have the default build targets that the users
of the library use that turn them off. 

That is largely my intention with removing -pendantic from GCC. The I still test MSVC and clang. I do test GCC as well, but the GCC versions on travis are so old that it is of no practical use.

derosier:
On Thu, Jul 11, 2019 at 10:28 AM MikkelFJ <notifications@github.com> wrote:

> Well, it does the job, but fails to illustrate the use of strncpy, which
> is apparently no longer part of the standard library. I'd much prefer if we
> could silence the GCC warning somehow.
>
>
?   AFAIK, strncpy is part of C18. At least it is as of the last publically
available draft as I didn't feel like purchasing a copy of the C18 standard
just to check.

It's also still in the ISO/IEC 9899:202x draft.

And - I'd say you're under no obligation to support a standard that hasn't
been ratified and published.

BTW, I just found this and thought it might help:
https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/

I don't know which warning it's hitting, but maybe one of these is the
right flag?

- Steve

mikkelfj:
> ?   AFAIK, strncpy is part of C18. At least it is as of the last publically
available draft as I didn't feel like purchasing a copy of the C18 standard
just to check.

It was just a passive aggressive annoyed statement reflecting on GCC preventing me from using a perfectly valid function. I didn't mean to say it is non-standard. Or if it is, it is an absurdly common non-standard function.

On your link - it seems to suggest it is part of -Wall, which we can't disable, so we need to find the specific flag. Maybe it is more aggressive in Fedora?

derosier:
On Thu, Jul 11, 2019 at 10:49 AM MikkelFJ <notifications@github.com> wrote:

> Well, I'm not building it and getting those errors, so I don't know
> which flags are the issue here. If the specific warning text from GCC
> was posted here, I'd be happy to give it a go.
>
> I'd appreciate if you can reproduce and possible find the proper flag til
> kill or set, -pendantic or otherwise.
>

I'm not the origional poster, nor do I use GCC 9, so can't help you there.



> The posting at the top of this issue mentions the following error for GCC
> 9:
>
> test/monster_test/monster_test.c:1948:5: error: ‘strncpy’ output truncated copying 5 bytes from a string of length 12
>
>  1948 |     strncpy(foobar->text, "hello, world", ns(FooBar_text_get_len()));
>
>
>
Perhaps: "-Wstringop-truncation"


> Note that the test specifically is testing truncation with strncpy: you
> have a long string, you want til fill the arrays as much as possible, and
> no more.
>
> My suggestion - choose a specific toolchain version you want to
> support.
>
> That is not really an option since a key selling point is that it runs
> everywhere, and it practically does.
>

Sure.  But it's not unreasonable for you to narrow your testing footprint
to a specific set. It is unreasonable to support (and by support, I mean
both build test and function test on each one) every version of every C
compiler ever made.  While the list isn't infinite, for all intents and
purposes it is. Your "customers" know that different compilers are
different and will throw different spurious crap out and is able to not
conflate a "we test on X" with "won't work on Y".  What I will guarantee is
that one of your users is going to pick this library up and try to use it
on some bizzare little compiler for some weird architecture that you never
knew existed and it's going to fail. You can't be responsible for that. And
you'd also not want to push in their funky workarounds that get it to work
on their platform but would break everyone else. Publishing the list just
lets people know what your boundaries and capabilities are in testing it.

- Steve

mikkelfj:
> But it's not unreasonable for you to narrow your testing footprint
to a specific set. 

Agreed. But it should work with warnings off most places - users are occasionally contributing to the portable library to help make that happen. Incidentally your comment about you focusing on a single environment on don't change tools often, is one reason there are so many old compilers to support, and then those new ones without a full stdlib.

derosier:
On Thu, Jul 11, 2019 at 11:06 AM MikkelFJ <notifications@github.com> wrote:
>
> ? AFAIK, strncpy is part of C18. At least it is as of the last publically
> available draft as I didn't feel like purchasing a copy of the C18 standard
> just to check.
>
> It was just a passive aggressive annoyed statement reflecting on GCC preventing me from using a perfectly valid function. I didn't mean to say it is non-standard. Or if it is, it is an absurdly common non-standard function.
>

Sorry, I misunderstood and took that literally.


> On your link - it seems to suggest it is part of -Wall, which we can't disable, so we need to find the specific flag. Maybe it is more aggressive in Fedora?
>

Try: -Wstringop-truncation

See also my followup post.

- Steve

mikkelfj:
@tonyhutter could you pleasy try to disable this warning: -Wstringop-truncation

derosier:
On Thu, Jul 11, 2019 at 11:09 AM MikkelFJ <notifications@github.com> wrote:

> But it's not unreasonable for you to narrow your testing footprint
> to a specific set.
>
> Agreed. But it should work with warnings off most places - users are
> occasionally contributing to the portable library to help make that happen.
> Incidentally your comment about you focusing on a single environment on
> don't change tools often, is one reason there are so many old compilers to
> support, and then those new ones without a full stdlib.
>
>
True and agreed.  And thus I don't tend to upgrade libraries on those
platforms either unless necessary. And even in cases where I need a fix
that's in a newer version, we'll tend to backport the single patch or the
minimal amount necessary to fix the issue. Because sometimes it's too much
work to move everything forward. Or the testing, validations, and
recertifications are cost-prohibitive. But it all depends on the project
and platforms involved.

- Steve

mikkelfj:
> Discussions about GCC aside, would you guys be ok with @derosier suggestion of simply disabling the warning around the strncpy() for the time being?
> 
> ```
> #pragma GCC diagnostic push
> #pragma GCC diagnostic ignored "-Wno-format-contains-nul" foo(b); /* no diagnostic for this one */
> #pragma GCC diagnostic pop
> ```
> 
> I'm just trying to reach a consensus on what I should do.

There is a [pwarnings.h](https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pwarnings.h) file that is mostly for MSVC because there are warnings that cannot be disabled in the build. The general principle is to disable warnings in the build when possible and necessary. Also because it is not necessarily included in all builds. Ideally a modern build only needs -std=c11 to work.

This is why I prefer this done in the CMake file.

The other question, should it be disabled? The warning you mention I suspect has to do with another print error - I think we should discuss that separately but I'm not sure what these flags do so ...

Anyway, lets fix strncpy first:

I think we should find and kill the warning, the only question is which one exactly and for which compiler. This is why a linked to the CMake file earlier. If any one of you can come up with a modified CMake file.

mikkelfj:
@tonyhutter and FYI: all flatcc generated files has en epilogue and prologue file included that deal with push and pop of warnings via some diagnostics files next to warnings file I linked to.  These are very general, such as static function not used. I'm not sure this captures test cases, but that could be fixed. But again, better fix special cases in CMake when possible.

mikkelfj:
BTW: in this long thread ...

memcpy is not acceptable because it is not robust against schema changes that makes the array longer or shorter. strncpy will truncated if necessary, or 0 pad, thus always creating a valid buffer entry.

mikkelfj:
Another option is to add the nonstring attribute to char arrays in the generated flatcc code. (See last paragraph of Steves linked article).
I'm not convinced yet. It would require ugliness in then generated code.

mikkelfj:
Could you test this branch:

https://github.com/dvidelabs/flatcc/tree/stringop-truncation

mikkelfj:
@tonyhutter I added your "fall through" changes to the new branch.

I also read through the `bfbs2json.c` and I don't see that there is any wrong null character so I'm inclined to leave the code as is - otherwise O needs to be renamed many more places, and it is not important. The fix with with null check is a loosing strategy as GCC will just attack another source line later.

Could you test if the `-Wno-format-contains-nul` solves the (hopefully) remaining issue in `bfbs2json.c`? Simply edit CMake similar to what I just did in the new branch.

tonyhutter:
@mikkelfj I just finished testing your stringop-truncation branch.  It correctly get rid of the semantics.c and monster_test.c build errors.   I was able to get bfbs2json.c to build with this patch:
```
diff --git a/samples/reflection/CMakeLists.txt b/samples/reflection/CMakeLists.txt
index db5b754..6a7dede 100644
--- a/samples/reflection/CMakeLists.txt
+++ b/samples/reflection/CMakeLists.txt
@@ -26,6 +26,14 @@ add_executable(bfbs2json bfbs2json.c)
 add_dependencies(bfbs2json gen_monster_bfbs)
 target_link_libraries(bfbs2json flatccrt)
 
+if (GCC_VERSION VERSION_GREATER_EQUAL 9.0)
+    # GCC 9 will warn if you try to printf a NULL.
+    # In https://github.com/dvidelabs/flatcc/pull/120 it was decided to disable
+    # this warning.
+    message(STATUS "Disabling GNU C compiler warnings: -Wno-format-overflow")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-format-overflow")
+endif()
+
 if (FLATCC_TEST)
     add_test(bfbs2json bfbs2json${CMAKE_EXECUTABLE_SUFFIX} ${GEN_DIR}/monster.bfbs)
 endif()
```
Do you want me to merge this patch, along with your stringop-truncation changes into my commit?

mikkelfj:
I think we should just add the warning to the one I wrote, since it likely originates from GCC 8 albeit GCC 8 might be less aggressive.

message(STATUS "Disabling GNU C 8+ string related compiler warnings")
...

I can either do this, or you can you make a new PR it you want. I'd squash merge it in.

tonyhutter:
@mikkelfj  I put  `-Wno-format-overflow` in `samples/reflection/CMakeLists.txt` so that it would be specific to `bfbs2json`.  If I add it to the toplevel CMakeLists.txt, it would be globally set.  Do you want it to be globally set?

mikkelfj:
I was wondering that myself / but I think there isn't currently any special flags, and I don't think it does any harm to have the setting globally. It does not force the flags onto end users.

CMakeLists.txt
```diff
@@ -193,6 +193,22 @@ elseif (CMAKE_C_COMPILER_ID STREQUAL "GNU")
...
+        if (GCC_VERSION VERSION_GREATER_EQUAL 8.0)
+            # Disable some GCC checks:
+            #
+            # -Wstringop-truncation:
+            # strncpy into char arrays in FlatBuffer structs, these are
+            # zero-paddded but not zero terminated
+            #
+            # -Wno-format-overflow:
+            # GCC 9 will warn if you try to printf a NULL.
+            #
```
mikkelfj: We are not trying to printf a NULL and doing so would be bad. It is just that gcc mistakenly think that we do.

```
 # -Wno-format-overflow:
 # GCC 9 warns on mistakenly assumed NULL character when
 # printing from a required FlatBuffer string field.
```

CMakeLists.txt
```diff
@@ -193,6 +193,22 @@ elseif (CMAKE_C_COMPILER_ID STREQUAL "GNU")
...
+            # It was decided these should be disabled globally in:
+            # https://github.com/dvidelabs/flatcc/pull/120
+            #
```
mikkelfj: No need to comment at that detail level here. There will be CHANGELOG entry.

tonyhutter:
@mikkelfj ok, my latest push updates the comments

mikkelfj:
OK thanks,
are we ready to merge?

tonyhutter:
Yep

PR #139, opened 2020-01-28
Custom assert definition for generated code

sciesielski:
It would be really nice to have a possibility to provide own assert function to generated code. In Google's flatbuffers it is done by providing FLATBUFFERS_ASSERT define. Following change introduces the same mechanism for flatcc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ executable also handle optional json parsing or printing in less than 2 us for a
   * [Docker](#docker)
   * [Cross-compilation](#cross-compilation)
   * [Custom Allocation](#custom-allocation)
+  * [Custom Asserts](#custom-asserts)
   * [Shared Libraries](#shared-libraries)
 * [Distribution](#distribution)
   * [Unix Files](#unix-files)
@@ -2248,6 +2249,30 @@ if doing so. There is a test case implementing a new emitter, and a
 custom allocator can be copied from the one embedded in the builder
 library source.
 
+
+### Custom Asserts
+
+On systems where the default POSIX `assert` call is unavailable, or when 
+a different assert behaviour is desirable, it is possible to override 
+the default behaviour in runtime part of flatcc library via logic defined 
+in [flatcc_assert.h](include/flatcc/flatcc_assert.h).
+
+By default Posix `assert` is beeing used. It can be changed by preprocessor definition:
+
+    -DFLATCC_ASSERT=own_assert
+    
+Please be aware of that, it only changes assert handling in flatcc library. For parsing 
+floating point numbers flatcc uses external Girsu3 library which also uses assertions. 
+To override asserts in Girsu3 library following definition can be used:
+
+    -DGRISU3_ASSERT=own_assert
+
+It is possible to disable assertions by using definitions:
+
+    -DFLATCC_NO_ASSERT
+    -DGRISU3_NO_ASSERT
+
+This will not affect inclusion of <assert.h> for static assertions.
 
 ### Shared Libraries
 
--- a/include/flatcc/flatcc_assert.h
+++ b/include/flatcc/flatcc_assert.h
@@ -1,0 +1,42 @@
+#ifndef FLATCC_ASSERT_H
+#define FLATCC_ASSERT_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+* These assert abstraction is __only__ for runtime libraries.
+*
+* The flatcc compiler uses Posix assert routines regardless
+* of how this file is configured.
+*
+* This header makes it possible to use systems where assert is not
+* valid to use.
+*
+* `FLATCC_ASSERT` is designed to handle errors which cannot be ignored
+* and could lead to crash.
+*
+* `FLATCC_ASSERT` can be overrided by preprocessor definition
+*
+* If `FLATCC_ASSERT` definition is not provided, posix assert is being used
+*/
+
+#ifdef FLATCC_NO_ASSERT 
+/* NOTE: This will not affect inclusion of <assert.h> for static assertions. */
+#undef FLATCC_ASSERT
+#define FLATCC_ASSERT(x) ((void)0)
+/* Grisu3 is used for floating point conversion in JSON processing. */
+#define GRISU3_NO_ASSERT
+#endif
+
+#ifndef FLATCC_ASSERT
+#include <assert.h>
+#define FLATCC_ASSERT assert
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* FLATCC_ASSERT_H */
--- a/include/flatcc/flatcc_builder.h
+++ b/include/flatcc/flatcc_builder.h
@@ -71,7 +71,7 @@ extern "C" {
 
 /* It is possible to enable logging here. */
 #ifndef FLATCC_BUILDER_ASSERT
-#define FLATCC_BUILDER_ASSERT(cond, reason) assert(cond)
+#define FLATCC_BUILDER_ASSERT(cond, reason) FLATCC_ASSERT(cond)
 #endif
 
 /*
--- a/include/flatcc/flatcc_flatbuffers.h
+++ b/include/flatcc/flatcc_flatbuffers.h
@@ -24,6 +24,7 @@ extern "C" {
 #include "flatcc/portable/pstdalign.h"
 
 #include "flatcc/flatcc_alloc.h"
+#include "flatcc/flatcc_assert.h"
 
 #define __FLATBUFFERS_PASTE2(a, b) a ## b
 #define __FLATBUFFERS_PASTE3(a, b, c) a ## b ## c
--- a/include/flatcc/flatcc_json_parser.h
+++ b/include/flatcc/flatcc_json_parser.h
@@ -14,7 +14,6 @@ extern "C" {
 
 #include <stdlib.h>
 #include <string.h>
-#include <assert.h>
 
 #include "flatcc/flatcc_rtconfig.h"
 #include "flatcc/flatcc_builder.h"
--- a/src/compiler/codegen_c_builder.c
+++ b/src/compiler/codegen_c_builder.c
@@ -501,7 +501,7 @@ int fb_gen_common_c_builder_header(fb_output_t *out)
         "static inline int N ## _start(NS ## builder_t *B)\\\n"
         "{ return flatcc_builder_start_table(B, K); }\\\n"
         "static inline N ## _ref_t N ## _end(NS ## builder_t *B)\\\n"
-        "{ assert(flatcc_builder_check_required(B, __ ## N ## _required,\\\n"
+        "{ FLATCC_ASSERT(flatcc_builder_check_required(B, __ ## N ## _required,\\\n"
         "  sizeof(__ ## N ## _required) / sizeof(__ ## N ## _required[0]) - 1));\\\n"
         "  return flatcc_builder_end_table(B); }\\\n"
         "__%sbuild_offset_vector(NS, N)\n"
--- a/src/compiler/codegen_c_reader.c
+++ b/src/compiler/codegen_c_reader.c
@@ -263,7 +263,7 @@ static void gen_union(fb_output_t *out)
         "{ return NS ## vec_len(uv__tmp.type); }\\\n"
         "static inline T ## _union_t T ## _union_vec_at(T ## _union_vec_t uv__tmp, size_t i__tmp)\\\n"
         "{ T ## _union_t u__tmp = { 0, 0 }; size_t n__tmp = NS ## vec_len(uv__tmp.type);\\\n"
-        "  assert(n__tmp > (i__tmp) && \"index out of range\"); u__tmp.type = uv__tmp.type[i__tmp];\\\n"
+        "  FLATCC_ASSERT(n__tmp > (i__tmp) && \"index out of range\"); u__tmp.type = uv__tmp.type[i__tmp];\\\n"
         "  /* Unknown type is treated as NONE for schema evolution. */\\\n"
         "  if (u__tmp.type == 0) return u__tmp;\\\n"
         "  u__tmp.value = NS ## generic_vec_at(uv__tmp.value, i__tmp); return u__tmp; }\\\n"
@@ -294,7 +294,7 @@ static void gen_union(fb_output_t *out)
         "static inline T ## _union_vec_t N ## _ ## NK ## _union(N ## _table_t t__tmp)\\\n"
         "{ T ## _union_vec_t uv__tmp; uv__tmp.type = N ## _ ## NK ## _type_get(t__tmp);\\\n"
         "  uv__tmp.value = N ## _ ## NK(t__tmp);\\\n"
-        "  assert(NS ## vec_len(uv__tmp.type) == NS ## vec_len(uv__tmp.value)\\\n"
+        "  FLATCC_ASSERT(NS ## vec_len(uv__tmp.type) == NS ## vec_len(uv__tmp.value)\\\n"
         "  && \"union vector type length mismatch\"); return uv__tmp; }\n",
         nsc);
 }
@@ -502,7 +502,7 @@ static void gen_helpers(fb_output_t *out)
         "#define __%sread_vt(ID, offset, t)\\\n"
         "%svoffset_t offset = 0;\\\n"
         "{   %svoffset_t id__tmp, *vt__tmp;\\\n"
-        "    assert(t != 0 && \"null pointer table access\");\\\n"
+        "    FLATCC_ASSERT(t != 0 && \"null pointer table access\");\\\n"
         "    id__tmp = ID;\\\n"
         "    vt__tmp = (%svoffset_t *)((uint8_t *)(t) -\\\n"
         "        __%ssoffset_read_from_pe(t));\\\n"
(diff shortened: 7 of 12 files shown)


derosier:
If you're going to add a user-visible feature, you need to add documentation too. I was looking at the patch and thinking "WTF? All this does is alias assert and nothing more." and took me a moment and a re-read of the pull request to understand why you're doing that. Not only do you need to add to the documentation, please document it inline as part of the commit message.

mikkelfj:
Please have a look at the verifier.c and builder.c logic - these both have customizable asserts. I'm not really sure why that is not the case in the code you are referencing, but it might be because it is documenting something that should never happen as opposed to potential user errors. Not saying that is the case.

sciesielski:
@mikkelfj I need custom assert handler for different build types, therefore I cannot use direstly assert from std lib. I would like to have some flexibility here. Just like I wrote, Google's flatbuffers helps me with that by providing such [mechanism](https://github.com/google/flatbuffers/blob/master/include/flatbuffers/base.h). As you can [see](https://github.com/google/flatbuffers/blob/master/include/flatbuffers/flatbuffers.h#L257), it is used across the code instead std assert. I cannot find any possibility to customize asserts in flatcc. Could you show me an example? Nevertheless I need to replace all occurrences of std lib assert with my custom implementation, without overriding assert itself, so I need it also in generated readers. 

@derosier could you tell me what kind of documentation is needed and where can I put it? I think it is not that complex change ;) It is not even documented in flatbuffers lib.

derosier:
On Tue, Jan 28, 2020 at 12:29 PM sciesielski <notifications@github.com> wrote:
>
> @derosier could you tell me what kind of documentation is needed and where can I put it? I think it is not that complex change ;) It is not even documented in flatbuffers lib.
>

What you need is what the feature is for and how to use it.
Preferably with an example. Which makes me think of it - you probably
should have a unit test that tests the feature now that it is to
become a supported feature.  In any case, back to the documentation, a
simple paragraph or two + a short example code snippet would be fine.
Obviously mention the define symbols it depends on, and how to get
them into your build (I'm guessing via a `gcc -D` flag or similar).
You could put it in the main README.md as that seems to be the bulk of
the documentation, or you could create a new MD file in the `doc/`
directory with the details.

- Steve

mikkelfj:
Here is an example in the builder logic. There is similar in the verifier. You are probably addressing reader logic where there might not be anything similar.

https://github.com/dvidelabs/flatcc/blob/f30b928f04b3c27552114f62dd7fe9433286a33f/src/runtime/builder.c#L31-L41

https://github.com/dvidelabs/flatcc/blob/f30b928f04b3c27552114f62dd7fe9433286a33f/src/runtime/builder.c#L307

Note that in the builder.c file, only some checks lead to a customized assert. There are other hardcoded assert statements. These hard asserts are used where errors are not expected.

As to documentation, I'm not sure the current assert overrides are particularly well documented but there is a section on debugging a buffer

https://github.com/dvidelabs/flatcc#debugging-a-buffer

WRT the current PR - I'm not sure what is the best approach because it is not just about overriding assert, but about enabling or disabling asserts in various scenarios. For example, assert when an incorrect call is made to the builder API in debug or release mode.

mikkelfj:
Also WRT the current PR - I'd like to avoid to have 10 different ways to define asserts, yet I also want to be able to control different aspects. Asserting on verify is different from asserting on build. Asserting on read is much more expensive than asserting on build.

sciesielski:
@mikkelfj check from your example looks more like soft asserts or debug logs. I want to have possibility to override hard assert with my own handler. Even if asserts on read shall never happen, they are there for a reason ;) If I have a policy in my project to not use assert, I'm a bit stuck because static code analyzers see usage of asserts in included flatcc headers. As you can see in Google's approach, without specifying own definition of FLATBUFFERS_ASSERT it still uses std lib assert by default. So it really makes it flexible. Together with FLATBUFFERS_ASSERT_INCLUDE of course, to allow usage of some defined interface for error handling.

pawelrutka:
I will bump this up. We also in project have same problem with asserts - they shall be replaced by some custom code when building for target. We use Flatbuffers C++ and this issue is resolved here, would be good to align because a lot of projects uses C & C++ API.
@mikkelfj woould be good to know how to approach with such change to make it happen. Thanks!

mikkelfj:
This is similar to how allocation has been customised. There is general file in include/flatcc for allocation, and it in turn references the portability layer in include/flatcc/portable. Some modules have their own alloc macro, but it is by default based on a common alloc macro as stated above.

https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_alloc.h

https://github.com/dvidelabs/flatcc/blob/f30b928f04b3c27552114f62dd7fe9433286a33f/include/flatcc/flatcc_builder.h#L190-L192


It seems that that current specialisations of ASSERT could continue as is, except to use FLATCC_ASSERT instead of assert, and in other places use FLATCC_ASSERT in place of "assert" directly.

I think this should be done by having a file "include/flatcc/flatcc_assert.h" which is included as "flatcc/flatcc_assert.h" or via "include/flatcc/flatcc_flatbuffers.h" as is done for alloc:

https://github.com/dvidelabs/flatcc/blob/f30b928f04b3c27552114f62dd7fe9433286a33f/include/flatcc/flatcc_flatbuffers.h#L26

Thus - the pull request would be much larger, address all uses of assert, and use FLATCC_ prefix instead of FLATBUFFERS_ prefix.

mikkelfj:
Also note the use of static_assert - this is very important to ensure that generated structs have the correct size in memory on any compiler the source is used with. A portable static_assert is hard to get right, but it has been done via 

https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pstatic_assert.h

and

https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pstatic_assert_scope.h

You might want to think about your requirements for static_assert as well!

sciesielski:
@mikkelfj I've updated code as you suggested
@derosier README.md has been updated too

include/flatcc/flatcc_assert.h
```diff
@@ -0,0 +1,39 @@
...
+* declaration
+*
+* If `FLATCC_ASSERT` definition is not providede, posix assert is being used
+*/
+
+#if !defined(FLATCC_ASSERT)
+#include <assert.h>
+#define FLATCC_ASSERT assert
+#elif defined(FLATCC_ASSERT_INCLUDE)
+     // Include file with forward declaration
```
mikkelfj: Please  use C style comments

mikkelfj: I'm not convinced that defining an include file is a good idea. It is not being used anywhere else, such as custom allocation. It's also possible to override https://github.com/dvidelabs/flatcc/blob/master/config/config.h

include/flatcc/reflection/flatbuffers_common_reader.h
```diff
@@ -14,7 +14,7 @@
 #define __flatbuffers_read_vt(ID, offset, t)\
```
mikkelfj: I'd rather have changes to generated code run as a separate commit. Did you makes these changes manually, or did you run the script in
https://github.com/dvidelabs/flatcc/tree/master/reflection
?

README.md
```diff
@@ -2249,6 +2250,25 @@ custom allocator can be copied from the one embedded in the builder
...
 
 
+### Custom Asserts
+
+In some system it would be not desirable to use Posix `assert`.
+Or it might, but more error handling is needed. In that case 
+[flatcc_assert.h](include/flatcc/flatcc_assert.h) can be used to override
+runtime assertion function. 
+
+By default Posix `assert` is beeing used. It can be changed by complie definition:
```
mikkelfj:
```suggestion
By default Posix `assert` is beeing used. It can be changed by preprocessor definition:
```

README.md
```diff
@@ -2249,6 +2250,25 @@ custom allocator can be copied from the one embedded in the builder
...
+
+    -DFLATCC_ASSERT=own_assert
+    
+If `own_assert` function is declared in separate header, it is possible to include 
+this header by using compile definition:
+
+    -DFLATCC_ASSERT_INCLUDE="own_assert.h"
+    
+Where `own_assert` can be reblaced by name of your own assert function/header name
```
mikkelfj:
```suggestion
where `own_assert` can be reblaced by name of your own assert function/header name.
```

README.md
```diff
@@ -2249,6 +2250,25 @@ custom allocator can be copied from the one embedded in the builder
...
+In some system it would be not desirable to use Posix `assert`.
+Or it might, but more error handling is needed. In that case 
+[flatcc_assert.h](include/flatcc/flatcc_assert.h) can be used to override
+runtime assertion function. 
+
+By default Posix `assert` is beeing used. It can be changed by complie definition:
+
+    -DFLATCC_ASSERT=own_assert
+    
+If `own_assert` function is declared in separate header, it is possible to include
```
mikkelfj:
```suggestion
If the `own_assert` function is declared in separate header, it is possible to include 
```

README.md
```diff
@@ -2249,6 +2250,25 @@ custom allocator can be copied from the one embedded in the builder
 library source.
 
 
+### Custom Asserts
+
```
mikkelfj: Suggest different wording:

On systems where the default POSIX `assert` call is unavailable, or when a different assert behaviour is desirable, it is possible to override the default behaviour via logic defined in [flatcc_assert.h](include/flatcc/flatcc_assert.h).

include/flatcc/flatcc_assert.h
sciesielski: Isn't that config.h used for compiler setup? Include file is needed for runtime part as custom assert can be overridden by some already declared function in existing header.

include/flatcc/reflection/flatbuffers_common_reader.h
sciesielski: reverted

include/flatcc/flatcc_assert.h
mikkelfj: Yes, I think you are right. Maybe there is something missing. I'm still not comfortable with controlling the build environment via defines in this manner, especially when it hasn't been necessary until know. You should be able to add your own files to build and use defines to take advantage of that implementation, which is why #ifndef is widely used. I'm open to suggestions.

README.md
```diff
@@ -2249,6 +2250,24 @@ custom allocator can be copied from the one embedded in the builder
...
+By default Posix `assert` is beeing used. It can be changed by preprocessor definition:
+
+    -DFLATCC_ASSERT=own_assert
+    
+If the `own_assert` function is declared in separate header, it is possible to include 
+this header by using preprocessor definition:
+
+    -DFLATCC_ASSERT_INCLUDE="own_assert.h"
+    
+where `own_assert` can be reblaced by name of your own assert function/header name
```
mikkelfj: Punctuation missing at the end.

include/flatcc/flatcc_assert.h
mikkelfj: Please have a look at https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_rtconfig.h

sciesielski: I think it can be done as it is. flatcc_alloc.h also has ifdefs for each functions to provide default implementation. In case of default include file it is strictly associated with default assert implementation which comes from posix. There is no point to have FLATCC_ASSERT_INCLUDE  defined in that case. If you would like to keep FLATCC_ASSERT_INCLUDE it would require to keep also an additional define flag to enable its usage.  flatcc_rtconfig.h seems to be not used in builder, or did I miss something?

mikkelfj: From memory, flatcc_rtconfig.h is only used for runtime, however, the compiler uses libflatcc which does include runtime components of libflatccrt, so indirectly it is used everywhere.

mikkelfj: If you are happy to remove the FLATCC_ASSERT_INCLUDE section, I think this PR is close to OK, and I can give it a closer look.

sciesielski: I need to inject somehow a header include in each place where custom assert is being used. Otherwise I'll end up with using undefined symbol. I see no other way to do that than by that define. In the same time, I do not want to add that include if I want to use posix assert. Adding #include only in case when FLATCC_ASSERT_INCLUDE is defined seems to be the only option to solve that.

mikkelfj: How about 
```
#ifdef FLATCC_CUSTOM_CONFIG
#include "flatcc_custom_config"
#endif

In flatcc/flatcc_rtconfig.h?

You can then add an include path and a custom header. If you do this, please check that it works with alt. assert.

Comment (sciesielski) [reply include/flatcc/flatcc_assert.h]: Sounds like lot of steps just to include one header. It would be hard to track all dependencies of headers inclusion.

Comment (mikkelfj) [reply include/flatcc/flatcc_assert.h]: But it will work with allocation etc. as well.

Comment (mikkelfj) [reply include/flatcc/flatcc_assert.h]: And, it is actually just two CFLAGS: -Imypath -DFLATCC_CUSTOM_RTCONFIG
You also change include path and have a verifier traces enabled etc.

Comment (mikkelfj) [reply include/flatcc/flatcc_assert.h]: Maybe a build option is also needed for custom include path.

Comment (sciesielski) [reply include/flatcc/flatcc_assert.h]: > Maybe a build option is also needed for custom include path.

which can be handled directly in flatcc_assert.h :) if you would like to have defined it in header, you can always add that define inflatcc/flatcc_rtconfig.h

I would add separate preprocessor option for alloc include path as usually memory related functions will be defined in other header than assert. 

Where in current implementation is desired place to define override alloc functions from flatcc_alloc.h? You have it ifdeffed just like I want to have that include for custom assert header, but I don't now where would you place that definitions if not as compile option.

Comment (mikkelfj) [reply include/flatcc/flatcc_assert.h]: there are endless such includes - alloc is just the most obvious example - that is why I dislike the approach.

Comment (sciesielski) [reply include/flatcc/flatcc_assert.h]: In README.md there is a statement: 

> while piloting a new platform
> flatcc_alloc.h can be used to override runtime allocation functions

. So I guess it is not intended to configure alloc in flatcc_rtconfig.h. I think it can also be done this way for assert. Flatcc_rtconfig.h seems to be used only in parser and verifier. I do not feel that confident to get  through all headers and dependencies and make it also working for builder and reader.

My only proposal is to keep that FLATCC_ASSERT_INCLUDE to have it aligned with Google's flatbuffers approach. If you are not interested, I can remove it and I'll add it in my local implementation.

Comment (mikkelfj) [reply include/flatcc/flatcc_assert.h]: Yes please remove. I think it is better for you to privately customise flatcc_assert.h in this case - which is also one reason to separate out the logic. I would like to make this simpler, but as I mentioned, adding special handling in build system for every imaginable configuration is not viable.

**Comment by mikkelfj** [comment]
Thanks, are you ready for merge? If so I'll give a closer look and merge if OK.

**Comment by sciesielski** [comment]
Yes, I have nothing else to add

**Comment by mikkelfj** [comment]
This looks ok, except the following the problem that is not easily resolved:

<assert.h> is included in
 include/flatcc/portable/grisu3*.h and include/flatcc/portable/pstatic_assert.h 

I already mentioned static asserts could be an issue. Grisu3 is used for fast numerical / text conversion and won't show up unless you work with JSON.

It is not possible to rely on FLATCC_ASSERT here because portable is a self-contained library.
It might be possible to define PORTABLE_NO_ASSERT or similar to disable use of asserts. Portable files also try to depend as little as possible on other portable files so they can be used in isolation, so making a framework for assert is not ideal either.

Grisu3 files are copied from external/grisu3 in the project root, so changes should originate there.

Changes to portable is best left for a separate commit.

**Comment by sciesielski** [comment]
I'm not sure if changing external lib code is a good idea. It can be hard to maintenance. 

What do you think about just informing about that issue with some additional action required to enable compilation in case someone want to override flatcc asserts and use girsu at once? It could look like this:

```C
#if !defined(FLATCC_ASSERT)
#include <assert.h>
#define FLATCC_ASSERT assert
#elif FLATCC_USE_GRISU3
#error "Grisu3 library uses posix assert. Please comment out this error if you want to override only flatcc asserts."
#endif
```

Regarding static assert I'm ok with that the header is included as long the `assert()` function is not called if FLATCC_ASSERT is defined

mikkelfj:
I maintain this grisu3 library after borrowing part of the C code from elsewhere, so it is OK to modify that library if necessary.

The grisu3 asserts can be changed to have a GRISU3_ASSERT, but not in this PR.
This needs to be documented.

mikkelfj:
I pushed changes to grisu3 code. If you merge that code and document that FLATCC_ASSERT will also need to override GRISU3_ASSERT when JSON parser is using it, it should be fine. Also document that assert is still included for static assertions. I suggest also adding

```
#ifdef FLATCC_NO_ASSERT 
#undef FLATCC_ASSERT
#define FLATCC_ASSERT(x) ((void)0)
/* Grisu3 is used for floating point conversion in JSON parsing. */
#define GRISU3_NO_ASSERT
#endif
```
before the `#ifdef FLATCC_ASSERT`. If you do this, please document.

mikkelfj:
In documentation, also note that FLATCC_ASSERT only affects the runtime library

mikkelfj:
Thanks. I'd change the formulation a bit in README and fix some typos on comments, but it is simpler that I fix that in a subsequent commit.

mikkelfj:
master branch now has updated doc and updated reflection source.

PR #143, opened 2020-05-12
Portability: fix ISO C99 unnamed struct/union

russdawson:
There is one unnamed union inside _**struct __flatcc_builder_frame**_ that, even when compiled with gcc _**-std=c99**_ flag, is only caught with the _**-pedantic-errors**_ flag.

This change-request simply names the union and updates instances where the _**frame()**_ macro accesses the union


mikkelfj:
thanks

mikkelfj:
@russdawson FYI: the pedantic option for GCC is no longer supported with GCC 8+ because it forces changes that would make it hard to maintain portable code.

Issue #111, opened 2019-03-22
Problems with compile option -Werror=undef


kristianQQQ:
I have created a schema and generated a header file. Trying to include this header in my project causes problems because the environment I work in forces -Werror=undef, and the flatcc headers in some places rely on undefined macros being zero.

`gcc test.c -Werror=undef -I flatcc\include`

```
flatcc/include/flatcc/portable/pdiagnostic.h:32:5: error: "PDIAGNOSTIC_AWARE_CLANG" is not defined [-Werror=undef]
 #if PDIAGNOSTIC_AWARE_CLANG
     ^
flatcc/include/flatcc/portable/pdiagnostic.h:41:5: error: "PDIAGNOSTIC_AWARE_MSVC" is not defined [-Werror=undef]
 #if PDIAGNOSTIC_AWARE_MSVC
     ^
flatcc/include/flatcc/portable/pdiagnostic.h:43:7: error: "PDIAGNOSTIC_AWARE_CLANG" is not defined [-Werror=undef]
 #elif PDIAGNOSTIC_AWARE_CLANG
       ^
flatcc/include/flatcc/portable/pdiagnostic.h:52:5: error: "PDIAGNOSTIC_AWARE_CLANG" is not defined [-Werror=undef]
 #if PDIAGNOSTIC_AWARE_CLANG

flatcc/include/flatcc/portable/paligned_alloc.h:66:32: error: "_XOPEN_SOURCE" is not defined [-Werror=undef]
 #if (defined _GNU_SOURCE) || ((_XOPEN_SOURCE + 0) >= 600) || (_POSIX_C_SOURCE + 0) >= 200112L
                                ^
flatcc/include/flatcc/portable/paligned_alloc.h:66:63: error: "_POSIX_C_SOURCE" is not defined [-Werror=undef]
 #if (defined _GNU_SOURCE) || ((_XOPEN_SOURCE + 0) >= 600) || (_POSIX_C_SOURCE + 0) >= 200112L
                                                               ^
flatcc/include/flatcc/portable/paligned_alloc.h:68:8: error: "__clang__" is not defined [-Werror=undef]
 #elif (__clang__)

flatcc/include/flatcc/portable/pdiagnostic_pop.h:4:5: error: "PDIAGNOSTIC_PUSHED_MSVC" is not defined [-Werror=undef]
 #if PDIAGNOSTIC_PUSHED_MSVC
     ^
[...]

PR #145, opened 2020-05-21
Some additional warning cleanup

mokus0:
I'm starting to pull flatcc into an existing codebase where I use a LOT of warnings-as-errors.  This cleans up/silences a few of the more aggressive ones.


mikkelfj:
I'll look into the other changes, but I am not happy about the signed conversion. There should not be such conversions in the code base, at least not without a cast. Where do you get such errors / warnings?

Also, what triggers all those warnings - i.e. what compiler and version do you use?
Some warnings are suppressed in the build, and if you use your own build you might not have the same settings. Specifically, GCC 8 has so many warnings that flatcc does not support pedantic warnings because some cannot be removed in the code base in a portable manner.

It can be OK to fix some warnings in the code base, like the -Wundef - I never got around to clean that up since there are many, and they depend on the compiler / platform as well - so just because you fix some for your build does not mean that they are gone. Therefore it is a non-trivial problem with risk of breaking things on other systems.

mokus0:
This project is using arm-none-eabi-gcc 8.2.1, the build downloaded from ARM.

The signed-conversion one looked a bit hairier than I wanted to get into, I'll go back and see exactly where they were.

mokus0:
And to further clarify - the build where I run into all of these is the build for my project using flatcc-generated code.  For compiling flatcc itself I have no problem using whatever settings are needed.

mikkelfj:
Thanks. I should look into that though it is strange that I don't get an error on other compilers. Actually there is a warning on mcvc 64 bit, not sure if it is the same: https://ci.appveyor.com/project/dvidelabs/flatcc/build/job/a21mx8vu70kcwb92/messages

Do you use pedantic warnings?

mokus0:
Pedantic is one of the relatively few things I don't use, I prefer to encourage use of widely-supported extensions when they help clarity

mokus0:
The specific list of -W flags my project is using, in case it helps:

```
-Werror=all
-Werror=extra
-Wfatal-errors
-Werror=conversion
-Werror=div-by-zero
-Werror=double-promotion
-Werror=endif-labels
-Werror=format
-Werror=implicit-function-declaration
-Werror=incompatible-pointer-types
-Werror=missing-braces
-Werror=missing-prototypes
-Werror=old-style-definition
-Werror=pointer-arith
-Werror=strict-prototypes
-Werror=undef
-Werror=write-strings
-Werror
```

mikkelfj:
FYI: the voffset_t type is defined in
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_types.h
and is unsigned 16-bit.

The warnings specific to GCC 8+
https://github.com/dvidelabs/flatcc/blob/e7b8ec22074dbe08343b4a4f84ef4cb96ffc58d3/CMakeLists.txt#L213-L227

You may also want to look into this issue that was closed with wont-fix regarding -Wundef:

https://github.com/dvidelabs/flatcc/issues/111

mokus0:
If the official solution is to suppress them myself, I can do so - just thought I'd offer up a more comprehensive fix.  I _believe_ that the way I changed it is strictly more standards-compliant, but as you say - with the huge variety of targets it's more or less impossible to know that it won't break something else.

mikkelfj:
Yes, I think it is OK to incrementally fix those warnings, but only if it can done at at low risk. I'll have a closer look. I do appreciate the contribution regardless.

mikkelfj:
You also want to ponder on the following comment in the CMake file:

https://github.com/dvidelabs/flatcc/blob/e7b8ec22074dbe08343b4a4f84ef4cb96ffc58d3/CMakeLists.txt#L175-L176

mikkelfj:
clang doesn't complain with strict-prototype, but perhaps the gcc problem is just the missing void in your commit. If so, we should add strict prototypes in the makefile and apply the fix. But I don't recall the details, so input is welcome.

mokus0:
I'm not sure about clang's behavior, but one guess is that it's treating it differently since that statement is serving as definition in addition to declaration, so it can clearly see that definition site has no arguments.  In GCC the missing `void` actually triggers two different warnings depending on whether it's a declaration (-Wstrict-prototypes) or definition (-Wold-style-definition), and in this case the one line will trigger both.

The main reason I like to use that warning in the first place is because without `void` in the argument list, ANSI C interprets the argument list very similarly to `(...)`, leading to the compiler erroneously accepting some incorrect usages of the function.  It's one mildly annoying difference from C++, but C++ at least accepts 'void' there as well.  I don't remember whether that's C++ standard compliant, but it's at least a widely supported behavior AFAIK.

include/flatcc/portable/pdiagnostic.h
```diff
@@ -20,12 +20,22 @@
 
 #if defined(_MSC_VER) && !defined(PDIAGNOSTIC_AWARE_MSVC)
 #define PDIAGNOSTIC_AWARE_MSVC 1
-#elif defined(__clang__) && !defined(PDIAGNOSTIC_AWARE_CLANG)
+#elif !defined(PDIAGNOSTIC_AWARE_MSVC)
+#define PDIAGNOSTIC_AWARE_MSVC 0
+#endif
+
+#if defined(__clang__) && !defined(PDIAGNOSTIC_AWARE_CLANG)
```
mikkelfj, line 27: Here you have the feature flag last unlike in paligned_alloc - I know this follows precedence in the current file, but we should probably choose one consistent approach.

include/flatcc/portable/pdiagnostic.h
```diff
@@ -20,12 +20,22 @@
...
 #define PDIAGNOSTIC_AWARE_CLANG 1
+#elif !defined(PDIAGNOSTIC_AWARE_CLANG)
+#define PDIAGNOSTIC_AWARE_CLANG 0
+#endif
+
 /* Can disable some warnings even if push is not available (gcc-4.2 vs gcc-4.7) */
-#elif ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 2)) && \
+#if ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 2)) && \
```
mikkelfj, line 34: This doesn't check if __GNUC__ is defined before using it. From memory, this is OK as long as it is not in an elif branch (for reasons beyond my comprehension), but we probably shouldn't rely on it if the goal is to clean this up.

include/flatcc/portable/pdiagnostic_pop.h
```diff
@@ -1,15 +1,25 @@
 #ifndef PDIAGNOSTIC_POP_H
 #define PDIAGNOSTIC_POP_H
 
+#if defined(PDIAGNOSTIC_PUSHED_MSVC)
```
mikkelfj, line 4: This is not necessary, because the flag is always defined in the corresponding push logic, and likewise for the following code.

include/flatcc/portable/pdiagnostic_pop.h
```diff
@@ -1,15 +1,25 @@
...
 #if PDIAGNOSTIC_PUSHED_MSVC
 #pragma warning( pop )
+#endif // PDIAGNOSTIC_PUSHED_MSVC
 #undef PDIAGNOSTIC_PUSHED_MSVC
```
mikkelfj, line 8: We should probably define it to 0 after undef so the multiple pops can rely on the flag being present. Likewise for subsequent code.

mikkelfj:
Could you please submit the prototype commit as a separate PR. It seems to be a clean solution and I have looked for other instances but it seemed like the only one. flatcc_accessors already added (void) to __size in a previous fix. You could also try to include strict prototypes for GCC in the CMakeLists.txt file, then the build on various GCC versions will take effect.

I added some review comments undef, it generally looks OK, but need some changes, see comments.

I'll try look for the signed conflict going forward.

mikkelfj:
I looked at the code where you got a sign conversion warning:

```
    if (__flatbuffers_voffset_read_from_pe(vt__tmp) >= sizeof(vt__tmp[0]) * (id__tmp + 3)) {\
        offset = __flatbuffers_voffset_read_from_pe(vt__tmp + id__tmp + 2);\
    }\
```
The problem is the multiplication, I presume: the code can be rewritten in types as:

```
    if (uint16_t >= size_t * (uint16_t + int-literal)) { ... }
```
What happens if you add cast, for example `(size_t)3` :
```
 if (__flatbuffers_voffset_read_from_pe(vt__tmp) >= sizeof(vt__tmp[0]) * (id__tmp + (size_t)3)) {...
```

I also wonder if the following line causes similar problems, but I suspect not, since it would break endless amounts of code?

` offset = __flatbuffers_voffset_read_from_pe(vt__tmp + id__tmp + 2);\`

I cannot test this myself since I don't readily have access to an obnoxious compiler.

mikkelfj:
I really don't want to disable signed warnings because there a plenty of places where the error could result in fatal errors. This is probably one more example of where GCC-8 has lost its way in overambitious warnings that achieve the opposite effect.

include/flatcc/portable/pdiagnostic.h
mokus0: It's actually not OK here, good catch. The reason it may not be caught is actually a different can of worms: clang defines __GNUC__.  The fact that it also checks the value of these flags will make it a bit uglier but I'll fix it.

include/flatcc/portable/pdiagnostic.h
mokus0: I have no preference; if it's styled differently elsewhere I can swap these around.

include/flatcc/portable/pdiagnostic_pop.h
mokus0: The if-defined checks you commented about were to take care of this, since the pre-existing code was using undef and I wanted to disrupt as little as possible.

I have no preference here either, but one thing that is potentially a little better about using undef is that after all the flatcc headers are done, these names will no longer exist in the preprocessor environment, which feels slightly cleaner to me.

mokus0:
I'll check out the signed warning issues.  If it's just the literal, that's an easy fix - there are unsigned literals, don't even need the verbose casting.  The `vt__tmp + id__tmp + 2` example would not trigger it because vt__tmp is a pointer, and + is left-associative so all intermediate expressions are just pointers.

mokus0:
Also - the reason your system probably doesn't complain about these conversions is because they are from `int` to `size_t`, which is larger on amd64.  If you compile for x86, it may complain on the same ones.

mokus0:
There are two other places my build trips over signed-conversion issues.  The first is pretty innocuous and easily fixed - "monster.fbs" defines a few fields as "short" but "monster.c" uses uint16_t.

The other is also easy to fix, but potentially changes generated code.  It's another case of `int` vs `size_t`, this time in some of the generated code - several macros in "flatbuffers_common_reader.h" create functions accepting `int n__tmp`, and then use them as `size_t`.

mokus0:
I took a look at adding -Wsign-conversion to the overall build, and that looks like it might be a much deeper rabbit hole than I can take the time for right now - the compiler has a huge number of them (almost 4000 in `scripts/build.sh` output - I imagine that's a lot of duplication of the same ones, but still pretty daunting).

I'll set up a PR to fix the ones required for my build, and then restructure this one to remove the -Wsign-conversion suppression.

mokus0:
This PR now only contains the -Wundef cleanup - there are now two others including cleanup for other warning types.

None of them are intended to claim to fully purge all warnings of that type.

include/flatcc/portable/pdiagnostic_pop.h
mikkelfj: yes, regarding the pop cleanup, that is is a valid point - but there are still other flags anyway.

include/flatcc/portable/pdiagnostic_push.h
```diff
@@ -29,12 +29,22 @@
...
+
+#if defined(__clang__)
 #pragma clang diagnostic push
 #define PDIAGNOSTIC_PUSHED_CLANG 1
-#elif ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+#else
+#define PDIAGNOSTIC_PUSHED_CLANG 0
+#endif
+
+#if ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
```
mikkelfj, line 43: This doesn't test if __GNUC__ is defined. 
(If it is, it is reasonably to assume __GNUC_MINOR__ is also defined.)

mikkelfj: Also, it changes behaviour because it is no longer an alternative to the clang branch. __GNUC__ is defined for clang.

mokus0: i'm not sure what happened here, this is not the code i thought I committed.

mokus0: ah, i see... i thought i had change this already because it's so similar to logic in pdiagnostic.h, and upon re-reading it now to figure out what's going on I was so blind to the difference I thought I was going insane.

include/flatcc/portable/pdiagnostic_pop.h
mokus0: It's not clear to me; do you want this changed?  I have no preference, but I've already spent far more time on all of this than I had intended, and I may soon just decide to hack around these issues in my own build.

mikkelfj: I't too tired to have an opnion right now though it would be nice if the pop logic could be more elegant / less bulky. Mostly I want consistency.

mokus0: No problem, I'm pretty tired myself.  If you decide there's a specific change you want me to make, I'm happy to make it, but I can't really justify (to my employer) taking too much additional time to restyle code.

mikkelfj:
Regarding ordering of variables on other code, let's leave it at that, and focus on push pop:

As I mentioned, the guard around each pop is not required with the new changes to push, however, I think it is fine to keep them. It adds some flexibility in not having to defined flags for all cases in push, should the need arise in the future.

In pdiagnostics.h the alternative 0 define for GCC is not using an elif unlike the other two cases, for consistency it would be noice to follow the same design with an elif branch.

Regardless, I think the commits are acceptable as is.

mokus0:
The difference in the GCC case is because I was trying to avoid the pattern `#if defined(X) && SOMETHING_USING_VALUE_OF(X)`.  Without doing both checks in the same expression, it makes it much harder to structure the check as a simple if/elif.  I have heard some preprocessors have issues with that, but not details about the specific issues.  I _suspect_ it's that -Wundef on some older versions of gcc may not be smart enough to recognize that the undefined symbol is only present in an unreachable branch of the expression.

I don't personally have any toolchain I need to support that requires this pattern, so if supporting it is not a concern I can restructure it to the simpler style used by the others.

mikkelfj:
Regarding elif I believe the problem you think of is that some preprocessors allowed undefined and test in the same if line, but does not allow for it in an elif branch. Or maybe that is just yet another oddity.

```C
#if !defined(PDIAGNOSTIC_AWARE_GCC) && defined(__GNUC__) && !defined(__clang__)
/* Can disable some warnings even if push is not available (gcc-4.2 vs gcc-4.7) */
#if (__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 2)
#define PDIAGNOSTIC_AWARE_GCC 1
#endif
#endif

#if !defined(PDIAGNOSTIC_AWARE_GCC)
#define PDIAGNOSTIC_AWARE_GCC 0
#endif
```

vs

```C
#if !defined(PDIAGNOSTIC_AWARE_GCC) && defined(__GNUC__) && !defined(__clang__)
/* Can disable some warnings even if push is not available (gcc-4.2 vs gcc-4.7) */
#if (__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 2)
#define PDIAGNOSTIC_AWARE_GCC 1
#endif
#elif !defined(PDIAGNOSTIC_AWARE_GCC)
#define PDIAGNOSTIC_AWARE_GCC 0
#endif
```

mikkelfj:
BTW: I't trying to get Travis included in the PR checks, any clue why only appveyor works?

mokus0:
The complexity I refer to is that in your second version, if `__GNUC__` is defined but less than 4, `PDIAGNOSTIC_AWARE_GCC` will end up unset.  Changing the `elif` branch into a second `if !defined` block allows it to ensure that that case will still lead to `PDIAGNOSTIC_AWARE_GCC` being set to 0.

Sorry, I don't know about the travis issue - it's been a long time since I've been active enough on github to have dealt with it much.

mikkelfj:
Yes, you are right, my mistake.
 If you have nothing to add, I'll merge.

BTW: I managed to get GCC 8 and 9 to build on Travis on the ci-more branch.

mokus0:
Nothing to add - I've been working on my protocol code itself and haven't run into any more warnings.

Thanks for the project, and for all the time you've spent reviewing!

mikkelfj:
You are welcome and thanks for contributing.

PR #147, opened 2020-05-26
Some sign conversion fixes

mokus0:
This does not fix all sign-conversion issues; there are still quite a few more throughout the code base.  It does fix a few that I've run into, though.


mikkelfj:
I haven't checked if changing int to size_t is the right thing to do, but if is, it should be changed in the code that generates the source, not in the generated source.

See https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/reflection/README

mokus0:
Ah, sorry - didn't realize that code was generated.

mikkelfj:
Yes the argument should have been size_t - it is the explicit length of a C string.

mokus0:
The update I pushed a few minutes ago moves all changes to the code generator and leaves the generated code untouched.

mikkelfj:
Good, I think it is fine, but I all review proper tomorrow (Europe).

PR #149, opened 2020-05-27
make input buffers const in *_create[_pe] generated from __*_build_vector_field

mokus0:
It looks like most places that generate "create" accessors already declare the source buffer `const`, but in this one spot it doesn't.

This appears to be a pretty low-risk change.  If my tracing through macros is correct, these accessors just immediately pass to more type-specific handlers which already have the buffer as `const`.


mokus0:
I see there's a test failure, but it looks like it may be spurious - from the log it appears the CI worker failed to check out the project.

mikkelfj:
The only concern is if the T type as macro argument conflicts with the const prefix for some types due to the ways macros work. I would also assume it is OK, but impossible to predict with macros.

I'm going to add it as is, since it seems to be the right thing to do. Thanks.

mikkelfj:
Yes, the AppVeyor builder is failing due to resources, not a concern.

mikkelfj:
I assume you run scripts/test.sh before pushing. Otherwise we will find out.

mikkelfj:
BTW: after I disabled the Travis check, it apparently started working. I am clueless.

mokus0:
Yes, I did run script/tests.sh before commit and it passed.  Thanks for the quick response!

mokus0:
BTW - regarding the concern about 'const' conflicting with expansion of 'T': that's one reason I have personally shifted to the style of placing const after identifiers.  It takes a bit of a mental reprogramming, but it's more compatible with the actual semantics of the language and it ends up making a lot of cases like that "just work".

Issue #151, opened 2020-06-08
A few -Wconversion and -Wsign-conversion warnings


mcrlc:
In the file `include/flatcc/flatcc_identifier.h` there are a few conversion issues. Since this header is included from `flatbuffers_common_reader.h` I couldn't build it in my project.
I used GCC 7.5 with the following flags (partial list) `-Wconversion -Wsign-conversion -Werror`.
I'll open a PR soon to point at the problems.

mikkelfj:
FYI: there were recently some fixes in this area:
https://github.com/dvidelabs/flatcc/pull/147

mcrlc:
Thanks. I actually looked at this PR before posting and it looks like it refers to different lines.
BTW: there are many more `-Wconversion` and `-Wsign-conversion` warnings, most of them I found in `builder.c`, but they look less trivial to fix. I might try some other time.

mikkelfj:
Yesterday I added `-Wsign-conversion` and `-Wconversion` to clang builds. Today I added `-Wsign-conversion` which required a lot more changes in addition to clang - it's broken for GCC 4.7 so disabled there (can't handle ternary operator). I also took your suggested cast from the PR since it wasn't a widespread problem. I tried to fix `-Wconversion` and did make some changes, but ultimately I stopped because `x = x + 1` or even `x = x + 1u` requires a cast, and `n = -n;` would also require a cast. That isn't reasonable.

I believe -Wsign-conversion is important in some cases, but it does leave a lot of ugly code that doesn't make it worthwhile in most cases. I might remove support for this warning in GCC in the future because GCC keeps getting more and more aggressive at the expense of code readability and portability. clang seems to strike a better balance although it also requires lots of casts.

Please let me know what you think. Changes are on master.

mikkelfj:
You can also see the build in the ci-more branch on travis - it tracks many more compiler versions but is only built when I explicitly push to that branch.

mcrlc:
Saw the changes. This actually fixed many of the other warnings we had. Your case about -Wsign-conversion is very reasonable, and I wish GCC didn't make it harder to use it.
Anyway, I'll close my PR now as it's no longer relevant. Thank you for maintaining this project!

mikkelfj:
Thanks, I'll be closing this issue. If you find some particular warning that you want fixed without being intrusive to the code, feel free to submit at new PR, but as discussed, the target is to support clang warnings not GCC.

PR #152, opened 2020-06-08
Fix some warnings on GCC 7.5

mcrlc:
When building with GCC 7.5 and the following flags `-Wsign-conversion -Wconversion` these lines produced warnings.
For some reason GCC didn't like the UL literal only in multiplications...
#151

### DIFF
--- a/include/flatcc/flatcc_identifier.h
+++ b/include/flatcc/flatcc_identifier.h
@@ -42,8 +42,8 @@ static inline flatbuffers_thash_t flatbuffers_type_hash_from_name(const char *na
 {
     uint32_t hash = 2166136261UL;
     while (*name) {
-        hash ^= (uint32_t)*name;
-        hash = hash * 16777619UL;
+        hash ^= (unsigned char)*name;
+        hash = hash * 16777619U;
         ++name;
     }
     if (hash == 0) {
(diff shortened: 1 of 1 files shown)


mikkelfj:
I don't understand why you can't cast char * to uint32_t - what does the error say?

As to the constant suffix, it is better to use `UINT32_C(value)` since U and UL are platform specific.

mikkelfj:
The header needs

```
#ifndef UINT8_MAX
#include <stdint.h>
#endif
```

in order to use UINT32_C, although it is already included via flatcc_flatbuffers.h via flatcc_types.h, but that depends on the order of include files.

mcrlc:
Regarding the `char` to `uint32_t` conversion, I actually found out an open bug in the GCC project for this type of conversions: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=49626
As for the literal suffix, I'll probably fix them tomorrow.

Thanks for the quick turnaround :)

mikkelfj:
I'll try make a commit based on clangs -Wsign-conversion warnings.
I'm not sure I want to deal with GCC bugs from 2011, that are still open. If you can't cast properly, all bets are off.

mcrlc:
I agree, it's probably safe to suppress this warning on GCC.

mikkelfj:
For reference, the proposed changes were committed directly to master branch along with many others. See also related issue #151.

PR #156, opened 2020-07-01
Fix standalone -c and --common_builder options

Elijan9:
"flatcc -c" without any source file doesn't produce the flatbuffers_common_builder.h header at the moment (line 430 cleared the cgen_common_builder option). Furthermore "flatcc --common_builder" mistakenly requires a source file (line 433), while it shouldn't need one. Both issues are addressed in this commit.


mikkelfj:
Ah thanks, I suspected this was a bit fuzzy, it gets confusing with all those flags.

PR #161, opened 2020-08-27
document fixes

yamt:


mikkelfj:
Thanks. Going forward it would be better to keep changes such as the one CMakeLists.txt separate, since it is almost a functional change.

yamt:
ok!

PR #163, opened 2020-09-17
Bugfix/iar toolchain warnings

amih90:


mikkelfj:
Thanks, while reviewing the rest, please note that the common builder code in reflection is generated and must be fixed in the code generator, not in the output: https://github.com/dvidelabs/flatcc/blob/master/src/compiler/codegen_c_builder.c

mikkelfj:
You can't use the UINT32_MAX to check for limits because the code is designed to be configurable to other sizes in other flatbuffer variants, like 64-bit offsets.

To see another example that attempts to work around compiler warnings in the same builder.c file:
```
    /*
     * That can happen on 32 bit systems when uoffset_t is defined as 64-bit.
     * `emit_front/back` captures overflow, but not if our size type wraps first.
     */
#if FLATBUFFERS_UOFFSET_MAX > SIZE_MAX
    check_error(vec_size < SIZE_MAX, 0, "vector larger than address space");
#endif
```

mikkelfj:
As to "Warning[Pe187]: use of = where == may have been intended"
I'm fine with the change, but note that other compilers generally accept a parentheses around an assignment to kill this kind of warning. As a consequence, you will likely find hundreds of other similar issues around in the code. Thus if you intend to fix this, you probably need to comb through the source. I'm not sure why you don't see more issues, but perhaps because you only compile generated code and not flatcc itself?

mikkelfj:
You can find more on the type macros here:
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/flatcc_types.h

mikkelfj:
And aside from the UINT32_MAX issue, I'd prefer if preprocessor guards cover a complete statement rather than part of an expression. I'd rather have duplicate code as long as it doesn't affect performance.

mikkelfj:
Closing due to lack of response.

PR #164, opened 2020-09-22
Update binary-format.md

t-rekttt:
I think you got a character missing at the place? It's small but caused me a bit of struggle while reading


mikkelfj:
Thanks, yes you are right.

PR #169, opened 2020-12-16
#168: add cmake helper function to generate flatbuffer code

### LINKED ISSUE #168: Add cmake helper function to simplify code generation
Add a cmake function that will take one or more flatbuffer definition files and an output directory. The function will then generate the C files in that directory.

This prevents code duplication in cmake files.

Pull request will follow.

wdobbe:
Resolves #168 .

Install a cmake module that provides cmake function `flatcc_generate_sources`.
This function simplifies generation of C code from flatbuffer definition file(s).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ executable also handle optional json parsing or printing in less than 2 us for a
 * [Online Forums](#online-forums)
 * [Introduction](#introduction)
 * [Project Details](#project-details)
+* [Use in CMake build](#use-in-cmake-build)
 * [Poll on Meson Build](#poll-on-meson-build)
 * [Reporting Bugs](#reporting-bugs)
 * [Status](#status)
@@ -186,6 +187,42 @@ their place.
 
 **NOTE: Big-endian platforms are only supported as of release 0.4.0.**
 
+## Use in CMake build
+
+If your project uses the CMake build system, you can include cmake module
+`cmake/FlatccGenerateSources.cmake` that provides the following cmake
+function:
+
+    flatcc_generate_sources(GENERATED_SOURCE_DIRECTORY <output directory>
+                            GENERATE_BUILDER
+                            GENERATE_VERIFIER
+                            EXPECTED_OUTPUT_FILES <list of expected output files>
+                            DEFINITION_FILES <list of flatbuffer files (.fbs)>
+)
+
+`GENERATE_BUILDER` and `GENERATE_VERIFIER` are boolean options. When specified
+they will instruct flatcc to generate builder / verifier source code.
+
+Optionally you can let cmake know the directory where the flatcc executable
+is located in environment variable `FLATCC_BUILD_BIN_PATH`. This is especially
+usefull when cross-compiling. In that case you should provide the directory
+where the build arch flatcc compiler executable is located.
+
+The flatcc_generate_sources function will create a cmake custom command to
+generate the output files during just before compilation (so in the build
+step, not the configure step).
+
+Example:
+
+    include(FlatccGenerateSources)
+    flatcc_generate_sources(GENERATED_SOURCE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/datadef
+                            GENERATE_BUILDER
+                            GENERATE_VERIFIER
+                            EXPECTED_OUTPUT_FILES datadef/seclif_protocol_reader.h
+                                                  datadef/seclif_protocol_builder.h
+                                                  datadef/seclif_protocol_verifier.h
+                            DEFINITION_FILES ${CMAKE_CURRENT_SOURCE_DIR}/datadef/seclif_protocol.fbs
+    )
 
 ## Poll on Meson Build
 
--- a/cmake/FlatccGenerateSources.cmake
+++ b/cmake/FlatccGenerateSources.cmake
@@ -1,0 +1,86 @@
+ cmake_minimum_required(VERSION 3.5)
+
+# Use the following function to generate C source files from flatbuffer definition files:
+#
+# flatcc_generate_sources(GENERATED_SOURCE_DIRECTORY <directory where to write source files>
+#                         GENERATE_BUILDER
+#                         GENERATE_VERIFIER
+#                         EXPECTED_OUTPUT_FILES <list of files that flatcc is supposed to generate>
+#                         DEFINITION_FILES <list of flatbuffer definition files (.fbs)>
+# )
+#
+# GENERATE_BUILDER and GENERATE_VERIFIER are boolean options. When specified they will instruct
+# flatcc to generate builder / verifier source code.
+#
+# With cross-compiling you should provide the directory where the flatcc compiler executable is located
+# in environment variable FLATCC_BUILD_BIN_PATH. If you use Conan and add flatcc as a build requirement
+# this will be done automatically.
+
+
+function(flatcc_generate_sources)
+
+    # parse function arguments
+    set(OUTPREFIX "FLATCC") #variables created by 'cmake_parse_arguments' will be prefixed with this
+    set(NO_VAL_ARGS GENERATE_BUILDER GENERATE_VERIFIER)
+    set(SINGLE_VAL_ARGS GENERATED_SOURCE_DIRECTORY)
+    set(MULTI_VAL_ARGS DEFINITION_FILES EXPECTED_OUTPUT_FILES CC_OPTIONS)
+
+    cmake_parse_arguments(${OUTPREFIX}
+                          "${NO_VAL_ARGS}"
+                          "${SINGLE_VAL_ARGS}"
+                          "${MULTI_VAL_ARGS}"
+                          ${ARGN}
+    )
+    if (GENERATED_SOURCE_DIRECTORY IN_LIST FLATCC_KEYWORDS_MISSING_VALUES)
+        message(FATAL_ERROR "No directory provided after GENERATED_SOURCE_DIRECTORY keyword")
+    endif()
+    if (NOT FLATCC_GENERATED_SOURCE_DIRECTORY)
+        set(FLATCC_GENERATED_SOURCE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+    endif()
+    message(STATUS "Flatcc sources will be generated in directory ${FLATCC_GENERATED_SOURCE_DIRECTORY}")
+
+    if (FLATCC_GENERATE_BUILDER)
+        list(APPEND FLATCC_CC_OPTIONS --builder)
+    endif()
+    if (FLATCC_GENERATE_VERIFIER)
+        list(APPEND FLATCC_CC_OPTIONS --verifier)
+    endif()
+
+    if (FLATCC_DEFINITION_FILES)
+        if (NOT EXISTS ${FLATCC_GENERATED_SOURCE_DIRECTORY})
+            file(MAKE_DIRECTORY ${FLATCC_GENERATED_SOURCE_DIRECTORY})
+        endif()
+
+        message(VERBOSE "Executing command ${FLATCC_COMPILER} ${FLATCC_CC_OPTIONS} -o ${FLATCC_GENERATED_SOURCE_DIRECTORY} ${FLATCC_DEFINITION_FILES}")
+        add_custom_command(OUTPUT ${FLATCC_EXPECTED_OUTPUT_FILES}
+                           COMMAND ${FLATCC_COMPILER} ${FLATCC_CC_OPTIONS} -o ${FLATCC_GENERATED_SOURCE_DIRECTORY} ${FLATCC_DEFINITION_FILES}
+                           WORKING_DIRECTORY ${FLATCC_GENERATED_SOURCE_DIRECTORY})
+    else()
+        message(WARNING "No flatbuffer definition files provided, no sources will be generated")
+    endif()
+
+endfunction()
+
+
+#### Main ####
+
+#When cross-compiling user can provide location of the flatbuffers to C compiler in build arch via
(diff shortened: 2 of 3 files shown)


cmake/FlatccGenerateSources.cmake
```diff
@@ -0,0 +1,86 @@
...
+        list(APPEND FLATCC_CC_OPTIONS --verifier)
+    endif()
+
+    if (FLATCC_DEFINITION_FILES)
+        if (NOT EXISTS ${FLATCC_GENERATED_SOURCE_DIRECTORY})
+            file(MAKE_DIRECTORY ${FLATCC_GENERATED_SOURCE_DIRECTORY})
+        endif()
+
+        message(VERBOSE "Executing command ${FLATCC_COMPILER} ${FLATCC_CC_OPTIONS} -o ${FLATCC_GENERATED_SOURCE_DIRECTORY} ${FLATCC_DEFINITION_FILES}")
+        add_custom_command(OUTPUT ${FLATCC_EXPECTED_OUTPUT_FILES}
```
madebr, line 55: This needs `DEPENDS` for dependency tracking of the input flatbuffer files.

cmake/FlatccGenerateSources.cmake
```diff
@@ -0,0 +1,86 @@
...
+    find_program(FLATCC_COMPILER flatcc
+                 PATHS ${FLATCC_BIN_PATH}
+                 NO_DEFAULT_PATH
+                 NO_SYSTEM_ENVIRONMENT_PATH
+                 NO_CMAKE_SYSTEM_PATH
+)
+else()
+    #Find compiler exe
+    find_program(FLATCC_COMPILER flatcc)
+endif()
```
madebr, line 81: I think it's ok to do this with one `find_program`:
- it's possible to override by doing `-DFLATCC_COMPILER=xxx`.
- Using `-DFLATCC_ROOT`, you can set a prefix of flatxx
- Using the `FLATCC_ROOT` environment variable, you can also set a prefix

I think the options `NO_DEFAULT_PATH`, `NO_SYSTEM_ENVIRONMENT_PATH` and `NO_CMAKE_SYSTEM_PATH` are not needed.

```suggestion
#user provided location where asn1c compiler executable is installed
find_program(FLATCC_COMPILER flatcc
     PATHS ${FLATCC_ROOT} ENV FLATCC_ROOT
     PATH_SUFFIXES bin
)
```

cmake/FlatccGenerateSources.cmake
```diff
@@ -0,0 +1,86 @@
+ cmake_minimum_required(VERSION 3.5)
```
madebr, line 1: I would split this file in 2:
- Export the `flatcc` target as `flatcc::flatcc` with `install(TARGETS)` and `install(EXPORT)`.
- Inside `flatcc-config.cmake`, do `include("${CMAKE_CURRENT_LIST_DIR}/FlatccGenerateSources.cmake")` which then only contains the function. Instead of the `FLATCC_COMPILER`, it can then use the imported `flatcc::flatcc` executable target.

cmake/FlatccGenerateSources.cmake
mikkelfj: Without reading carefully and from memory, just adding DEPENDS will break Ninja build because CMake doesn't handle dependecies properly, even if it works with Make. If you look at some of the examples that build with CMAKE, these are rather clumsy because of this.

madebr: But without DEPENDS, the target sources don't get regenerated when the input sources get modified?

madebr: [The `add_custom_command` documentation](https://cmake.org/cmake/help/latest/command/add_custom_command.html) does mention `ninja` with multiple arguments, but not with `DEPENDS` btw.

mikkelfj: I don't know because I am not a CMake expert. But I do know that I have spent several days trying to make it work (or something similar). Another who believed he could fix it had to admit he couldn't.

mikkelfj: I think DEPENDS is ok, I'm really don't recall the problem exactly. Here is the the monster test build:

https://github.com/dvidelabs/flatcc/blob/master/test/monster_test/CMakeLists.txt

mikkelfj: The problem I was having was that test cases cannot run properly with dependencies. I think I just disabled it with CMake as driver and ran tests with shell scripts.

mikkelfj:
Can you provide an example within FlatCC's test cases that uses this module?

I'm not convinced I want to accept this PR because I'd like to keep CMake maintenance to a minimum. It is has been rather painful.

cmake/FlatccGenerateSources.cmake
madebr: I'm looking into it now, I'll open a pr for installing `flatcc-config.cmake`.

This allows your users to use flatcc by doing `find_package(flatcc REQUIRED)` and call flatcc using `flatcc::flatcc`.

README.md
```diff
@@ -186,6 +187,42 @@ their place.
...
+                            EXPECTED_OUTPUT_FILES <list of expected output files>
+                            DEFINITION_FILES <list of flatbuffer files (.fbs)>
+)
+
+`GENERATE_BUILDER` and `GENERATE_VERIFIER` are boolean options. When specified
+they will instruct flatcc to generate builder / verifier source code.
+
+Optionally you can let cmake know the directory where the flatcc executable
+is located in environment variable `FLATCC_BUILD_BIN_PATH`. This is especially
+usefull when cross-compiling. In that case you should provide the directory
```
mikkelfj, line 208: useful

cmake/FlatccGenerateSources.cmake
madebr: It looks like on every `make`, all tests are re-built.
This is because the dependency tree is not correct.
I'll look into that too.

madebr: @wdobbe
You can then re-use/rebase your `FlatccGenerateSources.cmake` script on my pr.

mikkelfj (commented):
Regarding minimum required version, there is reason why flatcc uses cmake_minimum_required (VERSION 2.8)

Otherwise it will not build on some older conservative platforms. If it is really necessary to use a more recent version and it only affects users own programs opting in to this feature then maybe OK, but otherwise it would interfere with portability which is important to FlatCC.

cmake/FlatccGenerateSources.cmake
mikkelfj: #user provided location where asn1c compiler executable is installed

asn1c?

mikkelfj:
BTW: any idea why Travis build breaks on MacOS? Is it just Travis that hasn't upgraded its platform properly?

cmake/FlatccGenerateSources.cmake
mikkelfj: OK, here is an old discussion about broken dependencies - I haven't reread it, but linked here for reference:

https://github.com/dvidelabs/flatcc/pull/24#issuecomment-266718814

README.md
wdobbe: > Regarding minimum required version, there is reason why flatcc uses cmake_minimum_required (VERSION 2.8)
> 
> Otherwise it will not build on some older conservative platforms. If it is really necessary to use a more recent version and it only affects users own programs opting in to this feature then maybe OK, but otherwise it would interfere with portability which is important to FlatCC.

I had to update the cmake_minimum_required version from 2.8 to 3.1 in another recipe PR today because a conan-center hook rejects the PR if the cmake_minimum_required is not _at least_ 3.1.

In this case the minimum cmake version is 3.5 because from that version cmake function cmake_parse_arguments checks for value arguments that are miss their value, so it makes the function more fool-proof.

wdobbe:
> Can you provide an example within FlatCC's test cases that uses this module?
> 
This is how I call it in one of our company cmake files:

```
flatcc_generate_sources(GENERATED_SOURCE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/datadef
                        GENERATE_BUILDER
                        GENERATE_VERIFIER
                        EXPECTED_OUTPUT_FILES datadef/seclif_protocol_reader.h 
                                              datadef/seclif_protocol_builder.h 
                                              datadef/seclif_protocol_verifier.h
                        DEFINITION_FILES ${CMAKE_CURRENT_SOURCE_DIR}/datadef/seclif_protocol.fbs
)
```

> I'm not convinced I want to accept this PR because I'd like to keep CMake maintenance to a minimum. It is has been rather painful.

1. I'm sending the module upstream as well, so hopefully it will be in the next flatcc release anyway
2. The module is optional, the end-user can choose whether to use it or not.
3. I don't see the CMake maintenance problem. CMake has always been completely backwards-compatible, so once the function works it shouldn't require any maintenance.

cmake/FlatccGenerateSources.cmake
wdobbe: Sorry, copy/paste error from a similar script for the ASN1C compiler. 
I'll correct that.

mikkelfj:
I'm suggesting to add an example of using this module within the FlatCC project, but perhaps not replacing existing since it could break portability with CMake version. I'm not sure if the module only works after CMake install and the README should probably also clarify how to make it available.

README.md
wdobbe: > I'm suggesting to add an example of using this module within the FlatCC project, but perhaps not replacing existing since it could break portability with CMake version. I'm not sure if the module only works after CMake install and the README should probably also clarify how to make it available.

Ok, I will do that.
The only thing to do is that you have to do to make the module work is add the FlatCC cmake subdir to your CMAKE_MODULE_PATH.

mikkelfj: But if it works the same for correct arguments on earlier versions then an earlier MIN version  will still support the check for users with a recent CMake while not breaking the build for those who do not.

cmake/FlatccGenerateSources.cmake
wdobbe: > I'm looking into it now, I'll open a pr for installing `flatcc-config.cmake`.
> 
> This allows your users to use flatcc by doing `find_package(flatcc REQUIRED)` and call flatcc using `flatcc::flatcc`.

Yes, that is a good idea. I have plenty examples of how to generate the config files (correctly) so if you want me to do it, let me know.

cmake/FlatccGenerateSources.cmake
wdobbe: Yes, that would be a better solution.

wdobbe:
> BTW: any idea why Travis build breaks on MacOS? Is it just Travis that hasn't upgraded its platform properly?
When adding flatcc to conan-center-index we also had problems on MacOS because of a new security feature called Security Integrity Protection (SIP). That could be the case, however I currently don't see build logs for MacOS for this PR? And the only Mac I currently have access to (due to Covid-19 lockdown) is too old to have SIP.

wdobbe:
> I'm suggesting to add an example of using this module within the FlatCC project, but perhaps not replacing existing since it could break portability with CMake version. I'm not sure if the module only works after CMake install and the README should probably also clarify how to make it available.

Ok, I will do that once changes suggested by you and @madebr are implemented.

mikkelfj:
I'm developing flatcc on Mac Intel with Ninja. Travis has been building without issues before, except that it is often very slow. I think all travis builds use Make. It is unlikely that this PR triggered a build issue, but it is the first time that I see it.
I'm inclined to move everything (Windows, Mac, Linux) to GH Actions, but its not happening right away.

mikkelfj:
BTW: conan is different because it uses a binary dist, I think. Building from source should only give SIP issues if the core tools are broken (homebrew?, clang, make, cmake).

wdobbe:
This is the comment I added for the workaround at that time (see [here](https://github.com/conan-io/conan-center-index/blob/master/recipes/flatcc/all/test_package/conanfile.py) and [here](https://github.com/conan-io/conan-center-index/blob/master/recipes/flatcc/all/test_package/CMakeLists.txt)):

```
#On MacOS System Integrity Protection (SIP) will clear the DYLD_LIBRARY_PATH variable.
#As a result calling flatcc from cmake will currently not work if the flatcc executable
# is linked shared. As a workaround we generate the flatbuffer C files in the Conan recipe
# when on MacOS and flatcc option 'shared' is True.
```

So the issue I talked about only arises when the flatcc executable links dynamically to the flatcc library and flatcc is called from cmake.

mikkelfj:
DYLD, OK.

Build logs - follow details in the "Some checks were not successful"
https://travis-ci.org/github/dvidelabs/flatcc/jobs/750009859
The problem is homebrew trying to install ninja, it seems.
Nothing that I have changed in years.

wdobbe:
I'm not too familiar with homebrew (I use conan nowadays) however in the build logs I don't see anything go particularly wrong, apart from the timeout at the end. Maybe the homebrew install was stuck or too slow?

mikkelfj:
@wdobbe I don't have the full overview here, but it seems that @madebr has already integrated most or all of your changes into his PR #171 
If that PR passes review I'd probably squash it into master as a single commit. In that case you might not get listed as a contributor, you should probably make a commit to that PR. But as I said, I don't have the full overview here.

madebr:
I have used @wdobbe 's commit and kept his name, so he should be mentioned as co-creator.

mikkelfj:
Should we close this PR then?

mikkelfj:
Sorry for not following up on this. We have several build PRs going and I need to narrow this down. If you are still interested, please comment on PR #258. The plan is to make a release soon, then introduce changes to the build.

Issue #170, opened 2020-12-16
Migrate CI from Travis / Appveyor to Github Actions


mikkelfj:
It should not be too difficult to get a build working on MacOs, Linux, and Windows.
The problem is to get the build working with a long line of old compilers.

FlatCC has a default build and a separate ci-more branch that is merge into before a release. ci-more checks a fairly long list of old compiler versions. GH actions would need to have a script that installs these old compilers.

Travis is moving their CI to a new url and is currently broken on MacOS because something changed in the homebrew support.

Appveyor does not work with all MSVC compilers in the same build image so either old or new, which is not ideal.

And having multiple build systems is also not ideal.

If anyone wants to help out with this, it would be great. Otherwise this might take a while.

madebr:
See https://github.com/actions/virtual-environments for the available environments on github actions.
The windows images currently only support VS2017 (windows-2016) and VS2019 (windows-2019), with the possibility that windows-2016 goes away.

mikkelfj:
There are more, they are just not installed. MSVC can be pulled from somewhere and someone probably already made an action for that.

mikkelfj:
https://www.microsoft.com/en-us/download/details.aspx?id=5555
MSVC 2010 redist package < 5MB.
https://github.com/microsoft/setup-msbuild
https://github.com/ilammy/msvc-dev-cmd

Problem is, installing a compiler might conflict with an existing one. But Appveyor has many compilers on one image while Microsoft do not want to install old compilers on any image.

madebr:
While useful, I don't think these links provide a Microsoft compiler:

- A re-distributable only contains the dll's of microsoft's C(++) runtime, we need the c++ compiler
- microsoft/setup-msbuild is only about making available msbuild (=sln + vcxproj of Visual Studio)
- ilammy/msvc-dev-cmd is about easy access to a MSVC console (using vcvars). This needs a version of MSVC installed.

I found these older headless compiler packages:

- MIcrosoft Build Tools 2013: https://www.microsoft.com/en-us/download/details.aspx?id=40760: I'm not sure though it contains a c++ compiler. Is c++ categorized as managed?
- Visual Studio 2015: https://devblogs.microsoft.com/cppblog/announcing-visual-c-build-tools-2015-standalone-c-tools-for-build-environments/


VIsual Studio 2013 was the first Visual Studio version that was avaiable as a community edition.
[...]

PR #171, opened 2020-12-16
cmake: flatcc_generate_sources + add_custom_command + crossbuild + ci

madebr:
This pr does the following:
- use `add_custom_command` to generate the sources. Add all generated sources literally to `OUTPUTS`.
  The key to don't have sources recompile is to add the generated sources to `add_executable`.
The flatcc maybe has an argument to return what files are generated, but this is work for #169.
`flatcc` should have an option that returns the generated sources (and nothing else).
e.g.:
    ```
    flatcc -a monster.fbs
    flatbuffers_common_builder.h flatbuffers_common_reader.h monster_builder.h monster_reader.h monster_verifier.h
    ```
    This information + `flatcc -d monster.fbs` allows to know the dependant sources + generated sources.
- Create aliases (`flatcc::runtime`, `flatcc::libflatcc`, `flatcc::cli`) for the 2 libraries + executable, and use it everywhere. `make install` will also create these 3 targets. This means that the script by #169 can be used by flatcc itself + users of flatcc (by using imported targets). So `flatcc` can be used as:
    ```cmake
    find_package(flatcc REQUIRED)
    add_custom_command(OUTPUTS genrated.h
        COMMAND flatcc::cli input.fbs ...)
    add_executable(myexe source.c generated.h)
    target_link_libraries(myexe PRIVATE flatcc::runtime)
    ```

I have tested it on cmake 2.8.12.1 (with gmake and ninja).


mikkelfj:
Thanks. It sounds a bit of a hack to add generated files to executables, but if it works ...
Note that Windows builds are important, and Appveyor just failed.

Do you really need to all output files of flatcc all these places? I worked with a meson build where I just used the most important output file as a dependency because Ninja does not support multiple output files (there is an issue on it).

Note that flatcc supports generating gnu make style .dep files, but it probably isn't helpful here. I used it with the meson build where it worked fine.

On discussion in related pr on build dependencies. things are coming slowly back to me. I think the benchmarks tests cannot be built with CMake as it is, though I do not recall why. 

https://github.com/dvidelabs/flatcc/tree/master/test/benchmark

They are building on the old meson branch I once made:
https://github.com/dvidelabs/flatcc/blob/meson/test/benchmark/benchflatcc/meson.build

mikkelfj:
Can you please explain exactly what the problem is that you try solve:

You mentioned in other PR that tests were always building. But when I use ninja build incrementally, nothing gets build that isn't out of data, at least that I have noticed. Maybe this differ for Make?

madebr:
On current master, when you do:
```cmake
cmake .. -DFLATCC_TEST=ON
make
make install
```
The test executables are rebuilt on every make invocation
I didn't test the ninja behavior of current master.
Because `ninja` runs the compilation on one line, the rebuilds might be hidden for you?
What does `ninja -v` give?

I'm currently looking into the Windows failure,.
I can't reproduce it on my Windows box.

mikkelfj:
Right, I see that now. ninja -v -j1

mikkelfj:
You may also want to look into ci-more branch, especially on Windows to ensure that MSVC 2010 is supported.

test/json_test/CMakeLists.txt
```diff
@@ -1,43 +1,69 @@
...
+include_directories("${GEN_DIR}")
 
-add_custom_target(gen_monster_test_json ALL)
+# FIXME: possible improvement if cmake_minimum_required >= 3.1:
+#        use add_library(... INTERFACE) +  add_custom_command + target_sources(INTERFACE) + target_link_libraries()
+
+# Use add_custom_target + add_custom_command + add_dependencies because a parallel MSVC build will cause
+# add_custom_command to be run multiple times in parallel which causes errors opening file for write.
+# This problem does not pose when generated sources are used by only one target.
+add_custom_target (gen_json_test
```
madebr, line 17: @mikkelfj 
I suppose you don't want to have these sources hardcoded on master.
So I propose, if/when you accept these changes, to merge these into some `cmake_work` branch on which @wdobbe can base his pull request.
And only if everything falls together, you can merge into master.

test/json_test/CMakeLists.txt
```diff
@@ -8,6 +8,24 @@ set(DATA_SRC "${PROJECT_SOURCE_DIR}/test/flatc_compat")
 
 include_directories("${GEN_DIR}")
 
+# FIXME: possible improvement if cmake_minimum_required >= 3.1:
```
mikkelfj, line 11: I generally use # TODO: here

test/json_test/CMakeLists.txt
mikkelfj: I could make a branch, but I suppose @wdobbe might just as well use your branch directly from the PR?
I will probably squash it all before it lands.

madebr:
The error on appveyor was due to the interaction between a parallel build and cmake's `add_custom_command`.

On cmake, `add_custom_command` does not create a target (=`.vcxproj` project).
This causes the copy + flatcc to be called in parallel, which causes errors on Windows.

I have added a custom target in the json tests because that one is the only location where the sources generated by one `add_custom_command` are used by multiple `add_executable` targets.

This json test approach can be adopted by @wdobbe  's cmake script .
So all `add_custom_command`'s I have added should get replaced by his `flatcc_generate_sources`.

I also added a comment how this can be simplified if the cmake minimum version would get bumped to 3.1.
I have not verified this approach though.

Also note that cmake 3.19.1 emits the following warning on your cmake script:
```
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

test/json_test/CMakeLists.txt
```diff
@@ -1,43 +1,69 @@
...
 set(FBS_DIR "${PROJECT_SOURCE_DIR}/test/monster_test")
 
 set(DATA_DST "${CMAKE_CURRENT_BINARY_DIR}")
 set(DATA_SRC "${PROJECT_SOURCE_DIR}/test/flatc_compat")
 
-include_directories("${GEN_DIR}" "${INC_DIR}")
+include_directories("${GEN_DIR}")
 
-add_custom_target(gen_monster_test_json ALL)
+# FIXME: possible improvement if cmake_minimum_required >= 3.1:
```
madebr, line 11:
```suggestion
# TODO: possible improvement if cmake_minimum_required >= 3.1:
```

madebr:
@mikkelfj 

flatcc has a `-d` option to list all dependencies.
Does it have an option to list all generated sources?

mikkelfj:
> flatcc has a -d option to list all dependencies.
Does it have an option to list all generated sources?

No not that I can recall.

mikkelfj:
Well, the dependency file is actually a source to dependency list, but I think it is constrained because of Ninja. It's been a while. Like this:

ouput1 output2 : src1 src2 src3

or

ouput1 : src1 src2 src3
ouput2 : src1 src2 src3

madebr:
@mikkelfj 
I thought about it a bit further, and isn't useful for flatcc itself, because `flatcc` is not available while configuring cmake.

It is useful though for consumers of flatcc.
Using the following dependency information,
```
generated_header1 generated_header2: source.fbs
source.fbs: included1.fbs  included2.fbs
```
cmake can automatically deduce the dependency information.

mikkelfj:
Yes but it would not be good for the intended purpose because these are separate build steps, especially considering how ninja works and how build generators works towards it.

It would probably need to be a different switch.

The build model works like multiple sources to a single target in the build graph.
Technically there can be more output files, but the model treats them all as one. A single output file is enough to represent all of them. Theoretically the build could break halfway and produce an incomplete result, but ninja cannot handle multiple files currently.

Another option would be to ask flatcc to produce a flag file once it has successfully written all output files. But can also do that in a wrapper command: `flatcc ARGS && touch flagfile`

madebr:
I'm on Windows now, but I tested it with `cmake .. -GNinja` and it worked fine.
Even with adding multiple OUTPUTS to `add_custom_command`.
Perhaps a more recent ninja fixed this problem you're referring to?

Please also note that I have enabled `test/cgen_test`, which had a note about it not working with ninja.

mikkelfj:
Take a look at the meson build rules on the meson branch:

https://github.com/dvidelabs/flatcc/blob/meson/rules/meson.build

This build will detect if you make a change in a the third level include file and build a file that depends on any generated output, provided that file has set the fbs file as a dependency.

mikkelfj:
> Please also note that I have enabled test/cgen_test, which had a note about it not working with ninja.

Ahh, that was the one - it was four years ago ... thanks!

mikkelfj:
Ninja should have fixed this now:
https://github.com/ninja-build/ninja/issues/1184

mikkelfj:
I don't really have time to try understand the implications of this right now, but if flatcc does not already support the improved dep file format, I'll certainly be adding it at some point.

mikkelfj:
https://github.com/dvidelabs/flatcc/blob/ce931d0e34b9a4a9dd69939f2d5860c781b42d8b/src/compiler/flatcc.c#L209

madebr:
Ah, I misread the monster example.
I didn't see that it had all 3 `fbs` files as input.
So it is indeed possible to pre-calculate the expected outputs like you do in the meson branch.
I think this will be the easiest approach instead of adding a new flatcc option.

@wdobble I will take a go at it..

mikkelfj:
I'm not sure that I follow - where are the three input files listed? link?

flatcc has the -r (--recursive) option to generate everything included, or just the file listed.

madebr:
I was hinting at `test/json_test/CMakeLists.txt` and `test/monster_test/CMakeLists.txt`..
It has the main `monster_test.fbs` + all other included fbs sources listed in `DEPENDS`.

mikkelfj:
> DEPENDS flatcc_cli "${FBS_DIR}/monster_test.fbs" "${FBS_DIR}/include_test1.fbs" "${FBS_DIR}/include_test2.fbs"

Ah right. This is something I'd really like to avoid because in real life large projects this can become difficult to manage when schemas are modified and there are many included schema files. But you gotta do what you gotta do.

I'm not too keen on hacking a new option in flatcc either, but if it can work with a correct and improved depfile generation, I think that is the way to go. Only I'm not going to fix that right away. It also needs a backwards compatibility mode I suspect.

madebr:
@wdobbe 
I used your script as a base to generate the headers for the tests/samples.
The script is also installed.

@mikkelfj 
The output files and dependencies are extracted from the files using some regex'es.
Because travis-ci is stuck on Macos, I also included a github actions workflow that can replace appveyor and travis CI.

mikkelfj:
> The output files and dependencies are extracted from the files using some regex'es.

From which files? The dep files?

> Because travis-ci is stuck on Macos, I also included a github actions workflow that can replace appveyor and travis CI.

Much appreciated. I presume you have seen the new issue I raised: https://github.com/dvidelabs/flatcc/issues/170
Note that before I can switch from Travis/Appveyor it is important to support old compilers on Windows and Linux - but on a separate branch - it is too slow for master, at least on Travis / Appveyor. On GH Actions you can can make a branch test and use the same action script on multiple branches - in the current setup I use different scripts on ci-more.

mikkelfj:
Note deprec warning on CMake on MacOS build:

> CMake Deprecation Warning at CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

@wdobbe @madebr   
As to CMake version, I think the CMake version supported on Ubuntu 18.04 should suffice.
If I get this right that would be 3.10: https://packages.ubuntu.com/bionic/cmake

cmake/FlatccGenerateSources.cmake
```diff
@@ -1,86 +1,177 @@
...
+    message(VERBOSE "Flatcc output directory: ${FLATCC_OUTPUT_DIR}")
+    message(VERBOSE "Flatcc output files: ${OUTPUT_FILES}")
+    message(VERBOSE "Flatcc execute: flatcc;${FLATCC_COMPILE_FLAGS};${ABSOLUTE_DEFINITION_FILES}")
+    add_custom_command(OUTPUT ${OUTPUT_FILES}
+        COMMAND "${CMAKE_COMMAND}" -E make_directory "${FLATCC_OUTPUT_DIR}"
+        COMMAND flatcc::cli ${FLATCC_COMPILE_FLAGS} ${ABSOLUTE_DEFINITION_FILES}
+        DEPENDS flatcc::cli ${FLATCC_DEFINITION_FILES}
+    )
+    if(FLATCC_TARGET)
+        # FIXME: possible improvement if cmake_minimum_required >= 3.1:
```
mikkelfj: TODO
Or maybe we should switch to 3.1 now?

cmake/FlatccGenerateSources.cmake
```diff
@@ -0,0 +1,244 @@
+# Use the following function to generate C source files from flatbuffer definition files:
+#
+# flatcc_generate_sources(OUTPUT_DIR <directory where to write source files>
+#                         BUILDER
+#                         VERIFIER
+#                         DEFINITION_FILES <list of flatbuffer definition files (.fbs)>
+# )
```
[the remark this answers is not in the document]
wdobbe: Personally I have little experience with flatbuffers/flatcc. It is used by colleagues but at the moment I only work on the build system. The solution of @madebr obviously has more knowledge of which sources the flatcc compiler generates from an input file hence he could avoid the EXPECTED_OUTPUT_FILES.

wdobbe (changes requested):
Using flatcc::cli will fail I think when cross-compiling, because that target will then point to the host arch flatcc exe while we need the build arch flatcc exe.
I think it is unavoidable that the caller needs to provide the location of the build arch flatcc exe, preferably via an environment variable. In the Conan recipe for flatcc I already export such a variable when cross-compiling is detected. That variable is then automatically picked up be the flatcc_generate_sources function.

wdobbe:
By the way @madebr : thanks a lot for all the work you put into this. I really appreciate it.

README.md
```diff
@@ -186,6 +187,42 @@ their place.
...
+
+    include(FlatccGenerateSources)
+    flatcc_generate_sources(GENERATED_SOURCE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/datadef
+                            GENERATE_BUILDER
+                            GENERATE_VERIFIER
+                            EXPECTED_OUTPUT_FILES datadef/seclif_protocol_reader.h
+                                                  datadef/seclif_protocol_builder.h
+                                                  datadef/seclif_protocol_verifier.h
+                            DEFINITION_FILES ${CMAKE_CURRENT_SOURCE_DIR}/datadef/seclif_protocol.fbs
+    )
```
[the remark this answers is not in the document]
madebr: This options is already removed as the output files are calculated.
I forgot to update the readme.

test/optional_scalars_test/CMakeLists.txt
```diff
@@ -1,19 +1,14 @@
-include(CTest)
-
-set(INC_DIR "${PROJECT_SOURCE_DIR}/include")
 set(GEN_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
 set(FBS_DIR "${PROJECT_SOURCE_DIR}/test/optional_scalars_test")
 
-include_directories("${GEN_DIR}" "${INC_DIR}")
+include_directories("${GEN_DIR}")
```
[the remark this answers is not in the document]
madebr: In `src/runtime/CMakeLists.txt`, the following is done:
```cmake
target_include_directories(flatccrt PUBLIC $<INSTALL_INTERFACE:include> $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
```
This means that targets linked to `flatccrt` (or its alias `flatcc::runtime`) will inherit its include directories.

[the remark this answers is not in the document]
mikkelfj: OK, thanks

madebr:
> Note deprec warning on CMake on MacOS build:
> 
> > CMake Deprecation Warning at CMakeLists.txt:5 (cmake_minimum_required):
> > Compatibility with CMake < 2.8.12 will be removed from a future version of
> > CMake.
> 
> @wdobbe @madebr
> As to CMake version, I think the CMake version supported on Ubuntu 18.04 should suffice.
> If I get this right that would be 3.10: https://packages.ubuntu.com/bionic/cmake


It would be nice to bump to at least 3.1.
I think 3.1 suffices, but I didn't test it though.

This would allow the following usage:
```cmake
find_package(flatcc REQUIRED)

include(FlatccGenerateSources)
flatcc_generate_sources(flatcc_headers input.fbs ALL)

add_library(main main.cc)
target_link_libraries(main PRIVATE flatcc::runtime flatcc_headers)
```

madebr:
> Using flatcc::cli will fail I think when cross-compiling, because that target will then point to the host arch flatcc exe while we need the build arch flatcc exe.
> I think it is unavoidable that the caller needs to provide the location of the build arch flatcc exe, preferably via an environment variable. In the Conan recipe for flatcc I already export such a variable when cross-compiling is detected. That variable is then automatically picked up be the flatcc_generate_sources function.

@wdobbe 
I'm thinking about splitting up the compiler and runtime component.
Then you can specify the search path of the runtime/library.
See [the documentation of `find_package`](https://cmake.org/cmake/help/latest/command/find_package.html).
Using `<PackageName>_ROOT`, you can then override the location of the compiler.

I'm envisioning the following usage
```cmake
# Only import the runtime (flatcc::runtime and flatcc::libflatcc)
find_package(flatcc REQUIRED COMPONENTS runtime)
# Only import the compiler (flatcc::cli)
find_package(flatcc REQUIRED COMPONENTS compiler)
````

Internally, `Findflatcc.cmake` will do `find_package(flatccRuntime REQUIRED)` and `find_package(flatccCli REQUIRED)`.
So by settings `flatccCli_ROOT` to the root of a build prefix, and `flatccRuntime_ROOT` to the root of a host prefix, cross building is supported.
And the target `flatcc::cli` reamains available.

@mikkelfj 
By the way, are you happy with the names of the targets `flatcc::cli`, `flatcc::runtime` and `flatcc::libflatcc`?
These targets will be exposed to flatbuffer users.

**Comment by mikkelfj** [comment]
> By the way, are you happy with the names of the targets flatcc::cli, flatcc::runtime and flatcc::libflatcc?
These targets will be exposed to flatbuffer users.

I am not familiar with CMake namespaces / modules so I can't really say. But the final output products shouldn't change because that would break things.

I am a bit worried about libflatcc because already many users confuse libflatccrt with libflatcc and even if runtime is runtime, people tend to look for the word lib. So maybe flatcc::libflatccruntime ?

**Comment by madebr** [comment]
Namespaces are just targets with `::` that are aliases of other targets or are imported.

`flatcc::libflatccruntime` is a bit too much, imho.
How about the following?
`flatcc::cli`, `flatcc::runtime` and `flatcc::libcli`?

The outputs won't change. This is just about picking names to minimize confusion.

**Comment by mikkelfj** [comment]
How about flatcc:cli flatcc::compiler flatcc::runtime ? This matches the src directory layout.

**Comment by madebr** [comment]
I was confused by cli vs compiler, wouldn't people be confused as well?

**Comment by mikkelfj** [comment]
Probably. But I don't really see a solution that doesn't confuse.

**Comment by mikkelfj** [comment]
> Ideally, people should only use the new flatcc_generate_sources function + flatcc::runtime.

I wouldn't be using cmake, but in one of my own projects I use libflatcc to compile schema files into binary schema files which I then use to generate source code for other protocols.

**Comment by wdobbe** [comment]
> I'm envisioning the following usage
> 
> ```cmake
> # Only import the runtime (flatcc::runtime and flatcc::libflatcc)
> find_package(flatcc REQUIRED COMPONENTS runtime)
> # Only import the compiler (flatcc::cli)
> find_package(flatcc REQUIRED COMPONENTS compiler)
> ```
> 
> Internally, `Findflatcc.cmake` will do `find_package(flatccRuntime REQUIRED)` and `find_package(flatccCli REQUIRED)`.
> So by settings `flatccCli_ROOT` to the root of a build prefix, and `flatccRuntime_ROOT` to the root of a host prefix, cross building is supported.
> And the target `flatcc::cli` reamains available.

If that works its fine with me, but can we make that work as well within Conan when using the cmake_find_package generator? Because I suppose the flatcc config files will have to be removed in the Conan package when submitting to Conan-center-index.

File: cmake/FlatccGenerateSources.cmake [hunk 8]
```diff
@@ -0,0 +1,329 @@
...
+#   PATHS <include-path> [<include-path> [...]]
+#
+#     Add extra include search paths where flatcc should look for included defintion files.
+#
+#   COMPILE_FLAGS <flatcc-flag> [<flatcc-flag> [...]]
+#
+#     Add extra arguments to flatcc.
+#
+
+cmake_minimum_required(VERSION 3.3)
```
madebr, line 88: I had to bump the minimum required cmake version to 3.3, to enable this usage on MSVC:
```
include(FlatccGenerateSources)
flatcc_generate_sources(flatbuffer_headers DEFINITION_FILES input.fbs ALL)

add_executable(main main.cc)
target_link_libraries(main PRIVATE flatcc::runtime flatcc_generated::flatbuffer_headers)
```

madebr:
@wdobbe 
> If that works its fine with me, but can we make that work as well within Conan when using the cmake_find_package generator? Because I suppose the flatcc config files will have to be removed in the Conan package when submitting to Conan-center-index.

It is compatible with conan with minimal patching (at least, until conan can create IMPORTED executable targets).

@mikkelfj 
As I mentioned earlier, the minimum required cmake version for the `flatcc_generate_sources` function is 3.3.
I have also added a test (`test/install_test`) that tests an installed flatcc prefix.
See the last step of the github actions script.

CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-cmake_minimum_required (VERSION 3.1)
+# Ubuntu 14.04 (Trusty)
+#cmake_minimum_required (VERSION 2.8.12.2)
+# Centos 7
+#cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 2.8)
```
mikkelfj: This should be liftet to 2.8.12 because newer CMakes are deprecating older versions. See build log on GH Actions.

> CMake Deprecation Warning at CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

cmake/FlatccGenerateSources.cmake
mikkelfj: We should keep an example without flatcc_generate_sources and make a README comment about this in build section and about disabling tests for older CMake versions.

mikkelfj:
Looking at the monster sampe, there is FBS_DIR. But is not used directly and I don't think it is used by the generator.
So I suggest either using it explicitly in front of monster.fbs, or removing it. Same apply to other places where this might be the case. If FBS is needed it should be documented in README.

I didn't look closely, but I think the README section belongs under build section that is placed later. If more prominent position is needed in README for this or other build getting started doc, it should probably be a short reference to the relevant section. Open to suggestions.

wdobbe:
I sent PR madebr/flatcc#1 to @made_br's cmake_dep branch mostly to add a few more input validation checks in the flatcc_generate_sources function.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+  - push
+  - pull_request
+jobs:
+  test_unix:
+    name: 'Test Unix with CMake'
+    strategy:
+      matrix:
+        os: ['macos-latest', 'ubuntu-latest']
+        config: ['Release', 'Debug']
+        cmake_generator: ['Unix Makefiles', 'Ninja']
```
madebr, line 11: Do we still need to test Ninja?

cmake/FlatccGenerateSources.cmake
madebr: I have kept `test/monster_test_solo` as a `add_custom_command`, because it was using `--stdout`, which is not used by the `flatcc_generate_sources` function.
Is that enough?

.github/workflows/ci.yml
mikkelfj: Ninja definitely needs to be supported, but don't think Travis is testing it. I'm running it for all dev, so not critical. A single test build for single configuration on Linux would be nice though.

Somewhat (un)related: we need to make sure these changes work with MSVC 2010 and I don't think that happens with the current Appveyor build. If not, then some merge with a branch off of ci-more, one off tmp build, is needed for smoke testing. Once things work, ci-more will test any future changes before release builds. BTW: I like you suggestion of GHActions cron job for periodic ci-more tests.

cmake/FlatccGenerateSources.cmake
mikkelfj: Yes, but I suggest adding a note in the README.

mikkelfj:
On the syntax for generate_sources:

DEFINITION_FILES: suggest changing to SCHEMA_FILES.

And I'm not really sure what the current SCHEMA does but I think it is binary schema, in which case BINARY_SCHEMA or REFLECTION probably would be better. I'm not claiming the flatcc cli options necessarily make a lot of sense. Suggestions are welcome.

samples/monster/CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
...
-set(FBS_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+
+# Don't use FBS_DIR to test whether flatcc_generate_sources knows to use the correct file.
 
 flatcc_generate_sources(
     NAME monster_sample
-    DEFINITION_FILES monster.fbs
+    SCHEMA_FILES monster.fbs
     ALL
     OUTPUT_DIR "${GEN_DIR}"
```
mikkelfj, line 9: What happens if you also drop the OUTPUT_DIR ?

madebr: Then it defaults to `CMAKE_CURRENT_BINARY_DIR`.

mikkelfj: Would it make sense to change that to CMAKE_CURRENT_BINARY_DIR/generated ?

madebr: I think the choice of `generated` is too general.
Functions of cmake also choose `CMAKE_CURRENT_BINARY_DIR` by default.
e.g. https://cmake.org/cmake/help/latest/command/configure_file.html

madebr: The input schema files are also relative to CMAKE_CURRENT_SOURCE_DIR.

.github/workflows/ci.yml
madebr: I just ran `git diff origin/ci-more origin/master`.
The diff is rather long, shouldn't the only difference between these branches be `.travis.yml` and `appveyor.yml`?

samples/monster/CMakeLists.txt
mikkelfj: OK

mikkelfj: FYI: flatcc also has an input dir option, but I'm not sure it is worthwhile adding here.

.github/workflows/ci.yml
mikkelfj: Correct, but only after master is merge into ci-more. Before that, there is all the ongoing development, since I don't merge that often. Changes apply cleanly because changes only go into master, or the travis/appveyor files on ci-more.

mikkelfj: I can do a merge now if you want. Hang on ...

mikkelfj: pushed update. Travis dist updated to focal on ci-more (20.04 LTS). Not sure if it builds yet.

samples/monster/CMakeLists.txt
madebr: Is that option documented in `flatcc -h`?
I have added `-I`, which is needed to grep for included files for finding dependencies.

mikkelfj: -I is the option for input dirs.
flatcc -h :
`  -I<inpath>                 Search path for include files (multiple allowed)`

hmm maybe not for toplevel fbs files? Though I would think that also works.

madebr: Yes, `-I` is documented. I thought you were talking about another option.

.github/workflows/ci.yml
madebr: You can also test 2017 and 2019 by setting the `APPVEYOR_BUILD_WORKER_IMAGE` environment variable:
https://www.appveyor.com/docs/environment-variables/

samples/monster/CMakeLists.txt
mikkelfj: Yes, I'm not sure that is precise. But I don't remember. It is unlikely that the parser driver makes such a distinction.

.github/workflows/ci.yml
madebr: There is still a small difference between master and ci-more:
`include/flatcc/portable/piovec.h` and `include/flatcc/portable/pversion.h`.

samples/monster/CMakeLists.txt
madebr: Maybe it depends on the order of which the input files are given?
But I think not using `-I` is better.
Just imagine `gcc` looking in include dirs for a source file when you gave a wrong source as input.

.github/workflows/ci.yml
mikkelfj: huh? need to fix that, thanks.

samples/monster/CMakeLists.txt
mikkelfj: I assume you use gcc as an example and not as a related problem since flatcc -I paths should have nothing to do with gcc -I paths.

The flatcc -I option is intended for use when you have a library collection placed in some faraway dir and you do not want that mess in your schema files.

.github/workflows/ci.yml
mikkelfj: fixed on ci-more

samples/monster/CMakeLists.txt
madebr: Yes, as an example :smile:

.github/workflows/ci.yml
madebr: I have pushed a script that should (=untested) merge master into ci-more once every week.

mikkelfj: Great, thanks!

mikkelfj:
If there is an enhanced GH-actions ci-more build, the regular ci.yml should probably have a branch filter so it doesn't also run on ci-more.

madebr:
> If there is an enhanced GH-actions ci-more build, the regular ci.yml should probably have a branch filter so it doesn't also run on ci-more.

I've added the filter, although I don't think it is really needed.
The CI test is really short and fast.

madebr:
> On the syntax for generate_sources:
> 
> DEFINITION_FILES: suggest changing to SCHEMA_FILES.
> 
> And I'm not really sure what the current SCHEMA does but I think it is binary schema, in which case BINARY_SCHEMA or REFLECTION probably would be better. I'm not claiming the flatcc cli options necessarily make a lot of sense. Suggestions are welcome.

I've made this change.

Please check all options of the function, because once added (and released),
it's very hard (not to say impossible) to remove them.

.github/workflows/ci.yml
madebr: I've tested it a bit on a branch of mine and had to do some magic to make it work:

1. create a commit on ci-more with .travis.yml and appveyor.yml for master 
2. merge with master
3. revert .travis.yml and appveyor.yml
4. push ci-more (and trigger .travis.yml and appveyor.yml)

This means that every scheduled run of ci-more will create between 2 and 3 commits. The 2 extra commits are done to avoid merge conflicts around .travis.yml and appveyor.yml..

CMakeLists.txt
madebr: This is bumped to 2.8.12

.github/workflows/ci.yml
mikkelfj: If we remove travis and appveyor from master that won't be an issue, though it is also not a problem as is.
I'd like to completely get rid of travis because they are now changing their domain and the build is broken in strange ways recently. But that requires old compilers on GH Actions on Linux, ci-more.

.github/workflows/ci-more.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: 'Check whether the working directory is clean'
+        run: |
+          git diff --exit-code
+      - name: 'Check whether only .travis.yml and appveyor.yml differ with master'
+        run: |
+          if test $(git diff origin/master --name-only | wc -l) != 2; then
+            echo "ci-more should diverge from origin/master with exactly 2 files (.travis.yml and appveyor.yml)";
+            git diff origin/master --exit-code;  # origin/master
+            exit 1;
+          fi
```
madebr, line 49: When you end up removing `.travis.yml` and `appveyor.yml` from master, then `ci-more.yml` can be reduced.

```suggestion
      - name: 'Merge master into ci-more'
        run: |
          git merge origin/master -m "Merge master into ci-more"
      - name: 'Check whether the working directory is clean'
        run: |
          git diff --exit-code
      - name: 'Check whether only .travis.yml and appveyor.yml differ with master'
        run: |
          if test "$(git diff origin/master --name-only | xargs)" != ".travis.yml appveyor.yml"; then
              git diff origin/master --exit-code;
          fi
```

mikkelfj: > Please check all options of the function, because once added (and released),
it's very hard (not to say impossible) to remove them.

It's hard for me to tell if this is implemented correctly, so I am going (mostly) for the documentation. I really don't know how CMake operates in this context.

Note, I might deprecate --schema and rename to --binary-schema or --reflection in a future flatcc version, but that doesn't change much here.

Looking at the comments in FlatccGenerateSources.cmake:

I wrote earlier that I'm not sure that setting READER as the default argument. But it also seems that ALL, RECURSIVE is being set, which I think is more useful:

> set(output_options BINARY_SCHEMA COMMON COMMON_READER COMMON_BUILDER BUILDER READER VERIFIER JSON_PARSER JSON_PRINTER JSON)
    set(NO_VAL_ARGS ALL RECURSIVE ${output_options})
    set(SINGLE_VAL_ARGS NAME OUTPUT_DIR OUTFILE PREFIX)
    set(MULTI_VAL_ARGS SCHEMA_FILES COMPILE_FLAGS PATHS)
    
   But later the following section says differently:
   
   > if(default_option_reader)
        set(FLATCC_READER ON)
    endif()
    
    
   There is an old error message here, should be "No flatbuffer schema files provided":
   
   > if (NOT FLATCC_SCHEMA_FILES)
        message(FATAL_ERROR "No flatbuffer definition files provided")
    endif()


I don't understand this line:

>  set(MULTI_VAL_ARGS SCHEMA_FILES COMPILE_FLAGS PATHS)

Later you append -I prefix to FLATCC_PATHS - I'm note sure what a keyword args on other CMake context.

Regarding the COMPILE_FLAGS option, this is (probably) correct,  but I wonder if EXTRA_ARGS, EXTRA_FLAGS, or something else would be more meaningful.

madebr: > Note, I might deprecate --schema and rename to --binary-schema or --reflection in a future flatcc version, but that doesn't change much here.

The function now accepts `BINARY_SCHEMA` only.
This function/file is meant to be distributed along flatcc, so it can be kept in sync.
When some options changes in the future, there are ways to ease the transition.

> Looking at the comments in FlatccGenerateSources.cmake:
> 
> I wrote earlier that I'm not sure that setting READER as the default argument. But it also seems that ALL, RECURSIVE is being set, which I think is more useful:

Is `ALL` the new agreed-upon default for cmake then?
You can also choose for a non-standard option e.g. `COMMON` + `BUILDER` + `VERIFIER` + `READER` without `RECURSIVE`.
But I think this will be harder to document, for no gain.

> > set(output_options BINARY_SCHEMA COMMON COMMON_READER COMMON_BUILDER BUILDER READER VERIFIER JSON_PARSER JSON_PRINTER JSON)
> > set(NO_VAL_ARGS ALL RECURSIVE ${output_options})
> > set(SINGLE_VAL_ARGS NAME OUTPUT_DIR OUTFILE PREFIX)
> > set(MULTI_VAL_ARGS SCHEMA_FILES COMPILE_FLAGS PATHS)
> 
> But later the following section says differently:
> 
> > if(default_option_reader)
> > set(FLATCC_READER ON)
> > endif()

What this does, is iterate over all  output_options (`ALL BINARY_SCHEMA COMMON COMMON_READER COMMON_BUILDER BUILDER READER`) and if none of these are set, then assume the caller has passed `READER` (=default output option when no output has been selected).

> There is an old error message here, should be "No flatbuffer schema files provided":
> 
> > if (NOT FLATCC_SCHEMA_FILES)
> > message(FATAL_ERROR "No flatbuffer definition files provided")
> > endif()

Fixed (also removed all reference of definition file to schema file).


> I don't understand this line:
> 
> > set(MULTI_VAL_ARGS SCHEMA_FILES COMPILE_FLAGS PATHS)

This variable is later used in `cmake_parse_arguments`.
It is a list of arguments that except multiple arguments.
`SCHEMA_FILES` can be one file or more.

> Later you append -I prefix to FLATCC_PATHS - I'm note sure what a keyword args on other CMake context.
> 
> Regarding the COMPILE_FLAGS option, this is (probably) correct, but I wonder if EXTRA_ARGS, EXTRA_FLAGS, or something else would be more meaningful.

I copied the terminology from [`FindBISON`](https://cmake.org/cmake/help/latest/module/FindBISON.html). But I guess `EXTRA_FLAGS` is ok too.

wdobbe: I created another PR to @madebr 's branch.

Things still to do/check:
- When running cmake with BUILD_SHARED_LIBS=On the installed flatcc exe can't find the flatcc compiler shared lib (on Linux) unless you set LD_LIBRARY_PATH. I think we should add RPATH.
- Cross-compiling and Conan compatibility  to be checked

If you can postpone the merge to Tuesday or so, I can do that.

mikkelfj: @wdobbe

I'm in no hurry to merge.

@madebr

If the READER is default, what does the following mean?
> set(NO_VAL_ARGS ALL RECURSIVE ${output_options})

> Is ALL the new agreed-upon default for cmake then?

No I am raising the discussion. I would expect at least READER + COMMON_READER as default because without a COMMON_READER the result will not be usable unless the common file is created elsewhere.

But in general ALL + RECURSIVE will build anything a user would expect when not caring too much about arguments, which is why I think that will be a good choice. But I am open to suggestions.

It looks to me as if you expand all the -I  paths without the -I prefix and later try to parse it, but I am probable just not understanding the mechanism:

> set(MULTI_VAL_ARGS SCHEMA_FILES COMPILE_FLAGS PATHS)
> This variable is later used in cmake_parse_arguments.

mikkelfj:
BTW: ALL should imply recursive. flatcc cli documents -a as : -cwvr

mikkelfj:
Maybe it is right to use READER as default since flatcc documents this to be the default. It is easy to add ALL if that is needed.

cmake/FlatccGenerateSources.cmake
```diff
@@ -0,0 +1,340 @@
...
+    endforeach()
+
+    foreach(path ${FLATCC_PATHS})
+        # TODO: file(REAL_PATH) if cmake_minimum_required_version >= 3.19
+        if(NOT IS_ABSOLUTE "${path}")
+            set(path "${CMAKE_CURRENT_SOURCE_DIR}/${path}")
+        endif()
+        list(APPEND absolute_flatcc_paths "${path}")
+        list(APPEND FLATCC_ARGS -I "${path}")
+    endforeach()
```
madebr, line 149: In this loop, all `PATHS` (=include paths) are converted to absolute paths.
The absolute paths are appended to an internal variable + added to the arguments (with `-I` prepended to each path).

cmake/FlatccGenerateSources.cmake
```diff
@@ -0,0 +1,340 @@
...
+    endif()
+    if (FLATCC_ALL)
+        list(APPEND FLATCC_ARGS -a)
+        set(FLATCC_COMMON_BUILDER ON)
+        set(FLATCC_COMMON_READER ON)
+        set(FLATCC_BUILDER ON)
+        set(FLATCC_VERIFIER ON)
+        set(FLATCC_READER ON)
+        set(FLATCC_RECURSIVE ON)
+    endif()
```
madebr, line 221: `ALL` implies common (c), builder (w), verifier (v), reader and recursive (r).
Reader is built always.

madebr:
> Maybe it is right to use READER as default since flatcc documents this to be the default. It is easy to add ALL if that is needed.

My idea was to keep the differences in behavior between the function and flatcc as small as possible.

.github/workflows/ci-more.yml
madebr: > I created another PR to @madebr 's branch.

Thanks! And merged!

>     * When running cmake with BUILD_SHARED_LIBS=On the installed flatcc exe can't find the flatcc compiler shared lib (on Linux) unless you set LD_LIBRARY_PATH. I think we should add RPATH.

This is added, and tested.
See https://github.com/madebr/flatcc/actions/runs/432848623


>     * Cross-compiling and Conan compatibility  to be checked

I'll leave that up to you :D

mikkelfj:
@derosier Do you have any comments to the updated cmake setup?

samples/reflection/CMakeLists.txt
```diff
@@ -9,23 +7,19 @@ include(CTest)
...
+
 add_executable(bfbs2json bfbs2json.c)
-add_dependencies(bfbs2json gen_monster_bfbs)
-target_link_libraries(bfbs2json flatccrt)
+target_link_libraries(bfbs2json PRIVATE flatcc::runtime flatcc_generated::reflection_sample)
 
 if (FLATCC_TEST)
-    add_test(bfbs2json bfbs2json${CMAKE_EXECUTABLE_SUFFIX} ${GEN_DIR}/monster.bfbs)
+    add_test(NAME bfbs2json COMMAND bfbs2json "${GEN_DIR}/monster.bfbs")
 endif()
```
madebr, line 25: The flatcc samples and tests are only built when `FLATCC_TEST` is true.
So I think you can do `add_test` unconditionally.

test/monster_test_solo/CMakeLists.txt
```diff
@@ -1,21 +1,19 @@
...
+        "${FBS_DIR}/include_test2.fbs" "${FBS_DIR}/attributes.fbs"
 )
 
-include_directories("${GEN_DIR}" "${INC_DIR}")
-add_executable(monster_test_solo monster_test_solo.c)
-add_dependencies(monster_test_solo gen_monster_test_solo)
-target_link_libraries(monster_test_solo flatccrt)
+add_executable(monster_test_solo monster_test_solo.c "${GEN_DIR}/monster_test.h")
+target_include_directories(monster_test_solo PRIVATE "${GEN_DIR}")
+target_link_libraries(monster_test_solo flatcc::runtime)
```
madebr, line 17:
```suggestion
target_link_libraries(monster_test_solo PRIVATE flatcc::runtime)
```

test/cgen_test/CMakeLists.txt
```diff
@@ -1,43 +1,47 @@
...
+    target_include_directories(test_generated_objects PRIVATE "${PROJECT_SOURCE_DIR}/include")
+    set_target_properties(test_generated_objects PROPERTIES EXCLUDE_FROM_ALL ON)
+
+    add_custom_target(test_generated
+        COMMAND "${CMAKE_COMMAND}" -E remove -f "${CMAKE_CURRENT_BINARY_DIR}/test_generated${CMAKE_EXECUTABLE_SUFFIX}.c" $<TARGET_OBJECTS:test_generated_objects>
+        COMMAND "${CMAKE_COMMAND}" --build "${PROJECT_BINARY_DIR}" --target test_generated_objects --config $<CONFIG> --verbose
+    )
+
+    add_test(NAME test_generated COMMAND ${CMAKE_COMMAND} --build "${PROJECT_BINARY_DIR}" --target test_generated --config $<CONFIG>)
+endif()
```
madebr, line 47: @mikkelfj 
Looks like I forgot to insert a final newline in some final lines.
Would you be interested in [adding an editorconfig file](https://editorconfig.org/) to the repository?
It's a small file auto-configuring a lot of editors with basic formatting settings.

mikkelfj: I use editorconfig in other projects, so fine with me. tabs = 4 spaces.

samples/reflection/CMakeLists.txt
mikkelfj: Yes a noticed that as well, in various places all around.

test/cgen_test/CMakeLists.txt
madebr: Is it fine to use:
```
root = true

[*]
indent_style = space
indent_size = 4
```
Are there binary files committed?

mikkelfj: Not sure what root means.
Yes there are binary files committed. You might want to look for .gitattr or something.
Newline translation can also cause a problem in some text files due to diff in some test scripts - gitattr should fix that for windows - but this was done a long time ago.
You might want to make sure that the various shell script drivers keep working after these changes.
This includes benchmarks.

mikkelfj: BTW: and somewhat unrelated: I have found diff to be a fantastic test tool. Instead of creating hundreds of small test cases you  just have a larger test driver that produce output and you save and commit the output and diff against a test run. If you are friendly, you make sure to print some context in the output so the error can found when diff breaks.

mikkelfj: Notably the flatc compatibility test uses binary flatbuffers for reference testing.

mikkelfj: But to answer you question, I think the editorconfig is fine, as far as I can tell. There might be some special cases to handle as per the above.

madebr: The editor looks in all parent directories for `.editorconfig` files.
`root` instructs the editor to not look in parent directories.
This makes sure that a `.editorconfig` in e.g. a users' directory does not cause unwanted changes.

To avoid any unintended changes, I modified the `.editorconfig` to:
```
root = true

[{*.c, *.h, *.cpp, *.bfs, *.sh, *.cmake, *.cmake.in, *.txt}]
indent_style = space
indent_size = 4
insert_final_newline = true
trim_trailing_whitespace = true

[*.yml]
indent_style = space
indent_size = 2
insert_final_newline = true
trim_trailing_whitespace = true
```
I assume the editor will only apply these rules to text files, not binary files.
But I'm not sure.

mikkelfj: I think `*.json` should be added to `*.yml`. Are there any yml files?
And `*.fbs` to the source file list.

Maybe you right wrt. binary files, but otherwise, from memory:
`*.bin` and `*.bfbs` are binary files, the latter being binary schema files, for former typically unspecified flatbuffer files.

mikkelfj: Uh, and what are `*.bfs` ? if you meant `*.bfbs` they shouln't be there, as mentioned above.

mikkelfj: There are none of those bin extensions and the flatc_compat dir has some strange extensions for binary and json for historical reasons, important from the google repo. Nonetheless, flatcc outputs .bfbs files with the --schema option.

madebr: The `*.yml` files are the CI sources (appveyor + travis + github actions).
Yes, I meant to do `*.fbs`. Oops.

wdobbe: Hmm, setting the <packageName>_ROOT environment or cmake variable (that we use to find the build arch flatcc compiler when cross-compiling was introduced only with CMake 3.12 :-(.

Run `cmake --help-policy CMP0074` for more information.

So we have to increase cmake_minimum_required to 3.12 or modify the cmake config files.

wdobbe: And even when changing `cmake_minimum_required` to 3.12 I can't get cross-compiling to work. It looks like `find_package` is smart enough to skip packages that are not in the same architecture as the host architecture.

See following output when running cmake with --debug-find:

```
  find_package considered the following locations for the Config module:

    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/usr/flatccCliConfig.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/usr/flatcccli-config.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/flatccCliConfig.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/flatcccli-config.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/usr/flatccCliConfig.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/usr/flatcccli-config.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/flatccCliConfig.cmake
    /home/wdobbe/install/dynniq-yoctosdk/cortexa9hf/3.0.0/sysroots/cortexa9t2hf-neon-poky-linux-gnueabi/flatcccli-config.cmake
    /home/wdobbe/build/flatcc_x86_64/lib64/cmake/flatcccli/flatccCliConfig.cmake
    /home/wdobbe/build/flatcc_x86_64/lib64/cmake/flatcccli/flatcccli-config.cmake
    /home/wdobbe/build/flatcc_x86_64/flatccCliConfig.cmake
    /home/wdobbe/build/flatcc_x86_64/flatcccli-config.cmake
    /home/wdobbe/build/flatcc_cortexa9/flatccCliConfig.cmake
    /home/wdobbe/build/flatcc_cortexa9/flatcccli-config.cmake
    /home/wdobbe/build/flatcc_cortexa9/lib/cmake/flatcccli/flatccCliConfig.cmake
    /home/wdobbe/build/flatcc_cortexa9/lib/cmake/flatcccli/flatcccli-config.cmake

  The file was found at

    /home/wdobbe/build/flatcc_cortexa9/lib/cmake/flatcccli/flatcccli-config.cmake
```
find_package skips the flatccCli config file installed in /home/wdobbe/build/flatcc_x86_64/lib64/cmake/flatcccli/flatcccli-config.cmake and then continues and finally selects /home/wdobbe/build/flatcc_cortexa9/lib/cmake/flatcccli/flatcccli-config.cmake.

wdobbe: Maybe go back to the very first solution where I used an environment variable in function `flatcc_generate_sources` to override the flatcc cli executable ?

madebr: You need a higher minimum cmake version only when you're cross building.

madebr: I'll look into cross building tomorrow.
Maybe allow an overload e.g. `FLATCC_CLI` to be used as executable.

mikkelfj:
I'm not too happy about introducing a python dependency into the build.

madebr:
@mikkelfj 

The current cmake script supports reference files.
With `make regenerate_reference_files`, you can regenerate the reference files.
They are stored in `test/reference`.
The outputs of the tests are compared when running ctest.

This comparison is diabled for:
- `cgen_test` because it tests compiling a source
- `load_test` because its output contains variable data (bandwidth, throughput, time

It also uncovers a problem: the output of monster_test is not reproducible.
See https://github.com/madebr/flatcc/actions/runs/434787563 for the output.

madebr:
> I'm not too happy about introducing a python dependency into the build.

It's a fallback for when bash is missing.

madebr:
Indeed, this pr started as getting the dependencies in order calling flatcc in-build, and not it's also adding reference files to the test system.
I'll remove the last commits, and open an issue.

mikkelfj:
> It's a fallback for when bash is missing.

We should try to avoid such dependencies. There are driver scripts now, but you are not forced to use them, except the benchmarks, and that isn't ideal either.

Where diff is needed, there is an equivalent for Windows using .bat files which I have used elsewhere, though I'm not convinced it is necessary to do that here.

mikkelfj:
It is called fc.exe for file compare, and not "fc" because that means something else in cmd.exe interpreter.

madebr:
Ok, I'll remove the feedback.
It looks like github actions and appveyor have bash + diff.
So python is not tested either way.
Appveyor is giving problems with (I think) endlines (`\r\n` vs `\n`).

madebr:
Btw, the reference files are now removed from this pr.
I'll open a pr when this one is finished.

mikkelfj:
Windows has bash and some other unix tools wherever there is git, but GH actions isn't the only environment, hence best to keep dependencies to a minimum.

madebr:
Remind me if I've forgotten to add `fc.exe` support on Windows when that pr gets opened. :smile:

mikkelfj:
But I still don't think it is necessary to rework all the tests. If I were rewriting all the JSON tests today, I would consider it. I use diff to test scripting languages because there are so many weird expressions to test that step by step test cases would explode.

madebr:
You don't need to rewrite any test?
What it does is capture the stdout/stderr of every test, and compare it to reference files.
These reference files can be generated by some make target.

mikkelfj:
The one test that would benefit from diff, is the one with the little whale. The cgen test. Currently it only tests that it is running, not that it is running correctly. And this is exactly a scripting language (or schema language) test.

mikkelfj:
Well, let's take that in a separate PR.

madebr:
Ok, can you look at #172?
That is the only diff that I could detect between runs on my pc and on runs on CI.

wdobbe:
I created a new PR to @madebr 's cmake_dep branch.
It makes cross-compilation work.

Also updated the README.md: 
- moved the cmake build info further down as requested by @mikkelfj 
- updated the section about cross compilation

mikkelfj:
@madebr if you merge master, travis should build again.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+          cd build_native
+          cmake .. -DBUILD_SHARED_LIBS=${{ matrix.shared_build }} -DCMAKE_BUILD_TYPE=${{ matrix.config_build }} \
+              -DFLATCC_INSTALL=ON -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/prefix_build
+          cmake --build . --parallel
+          cmake --build . --target install
+      - name: 'CMake configure/build/install host flatcc (cross)'
+        run: |
+          mkdir build_mingw
+          cd build_mingw
+          cmake .. -DCMAKE_TOOLCHAIN_FILE=../cmake/x86_64-w64-mingw32-toolchain.cmake -DFLATCC_ALLOW_WERROR=OFF \
```
madebr, line 133: I had to disable `-Werror` for mingw because there were `-Wformat` warnings when building for mingw.
See https://github.com/madebr/flatcc/runs/1591429817?check_suite_focus=true for the output.

mikkelfj: That's no good. And there is actually a problem in the source code.

` return (size_t)sprintf(literal, "%s(%llu)", cast, llu(value->u));`

The should be %PRIu64 or something, not llu. That is old.

More generally, if there is a really good reason, the main CMakeList.txt file disables specific warning types for specific compiler versions, or the warnings are disabled in a file always included in the build, but that is mostly when compilers consistently breaks standards compliant code, not mentioning any names, but there are some MSVC warnings there: https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pwarnings.h

mikkelfj: fixed on master, please try with warnings.

madebr: See https://github.com/madebr/flatcc/runs/1591699375?check_suite_focus=true

mikkelfj: This is more or less unfixable from a quick internet search. Mingw32 does not like I64u I64d modifiers which it itself defines PRI macros to in inttypes.h, but it also doesn't like %llu and %lld. Compiling runtime only might solve it.
I'd patch inttypes.h if I could, the machinery is there, but the solution is not, it seems.

mikkelfj: If not already applied, could you try one or several of
-DFLATCC_PORTABLE
-std=c99
-std=c11

madebr: I've applied the suggestion from https://stackoverflow.com/questions/44382862/how-to-printf-a-size-t-without-warning-in-mingw-w64-gcc-7-1:
`-Wno-pedantic-ms-format`

When building local, the error moves to `reflection_test`, which contains another `%lld`.

mikkelfj: My awesome ag search powers have failed me.

madebr:
> I wonder if this flag belongs in the main CMakeList.txt file since you could compile mingw natively on Windows too?

Let's keep in the back of our head when/if you test it on mingw@WIndows?

mikkelfj:
>Let's keep in the back of our head when/if you test it on mingw@WIndows?

right, good find btw.

.github/workflows/ci.yml
mikkelfj: Fixed in master

madebr: YES!
It's all-green once again!
https://github.com/madebr/flatcc/actions/runs/436827350

mikkelfj:
In the mingw cross build:

>-- Installing: /home/runner/work/flatcc/flatcc/prefix_host/lib/libflatccrt.dll.a
-- Installing: /home/runner/work/flatcc/flatcc/prefix_host/bin/libflatccrt.dll
-- Installing: /home/runner/work/flatcc/flatcc/prefix_host/lib/libflatcc.dll.a
-- Installing: /home/runner/work/flatcc/flatcc/prefix_host/bin/libflatcc.dll
-- Installing: /home/runner/work/flatcc/flatcc/prefix_host/bin/flatcc.exe

is dll.a a new word for .lib?

.github/workflows/ci.yml
mikkelfj: 👍

madebr:
On Windows, shared libraries (dll's) have an import library.
The import library contains the exported symbols, users of the library can use.

MSVC default to use `.lib` for static libraries, and GNU `.a`.
The convention for mingw is to use `.dll.a` for these import libraries.
Thay way, simultaneous builds of static and shared builds (which is the default of libtool) don't clash.

mikkelfj:
> That way, simultaneous builds of static and shared builds (which is the default of libtool) don't clash.

OK

madebr:
> > That way, simultaneous builds of static and shared builds (which is the default of libtool) don't clash.
> 
> OK

Note that using `-lsome_library` will match both `libsome_library.dll.a` and `libsome_library.a`.
I think the linker prefers the import library, but I'm not sure. I guess using `-Bshared` and `-Bstatic`, you can select between the two.

madebr:
> > Let's keep in the back of our head when/if you test it on mingw@WIndows?
> 
> right, good find btw.

So,
gcc mingw 8 from chocolatey needs the workaround:
https://github.com/madebr/flatcc/runs/1592328584?check_suite_focus=true

gcc mingw 10 from msys2 does not need the workaround:
https://github.com/madebr/flatcc/runs/1592230351?check_suite_focus=true

The proper way to add this flag is to do a [`try_compile`](https://cmake.org/cmake/help/latest/command/try_compile.html).
Or the simpler way is to just add the flag when [`MINGW`](https://cmake.org/cmake/help/latest/variable/MINGW.html) is true-ish.

prince-chrismc:
If this gets merged the would would benefit from a release ❤️

mikkelfj:
> The proper way to add this flag is to do a try_compile.
Or the simpler way is to just add the flag when MINGW is true-ish.

Adding the flag is the right way. No need to not set it on more recent compilers since no-one in their right mind would add MSVC specific formatters to the source tree outside of PRI macros.

FlatCC has not had the need of a single try_compile despite being used on rather different platforms, and I'd like to keep it that way.

mikkelfj:
@prince-chrismc 
> If this gets merged the would would benefit from a release ❤️

Yes, it is about time for a release, also because of new optional scalar feature. But I try to not release immediately after making larger changes.

scripts/cleanall.sh
```diff
@@ -11,6 +11,9 @@ rm -rf release
 rm -f bin/flatcc*
 rm -f bin/bfbs2json*
 rm -f lib/libflatcc*
+rm -f bin/*test*
+rm -f bin/flatc_compat*
+rm -f bin/monster*
 if [ -d bin ]; then
```
mikkelfj, line 17: Why do these tests end up here in the first place? That was never the intention with the bin dir.

madebr: When building shared libraries on Windows, the tests failed because they could not find `libflatccruntime.dll`.
Maybe this can be revisited by setting the PATH environment variable to `$prefix/lib`.

mikkelfj: Wouldn't it make more sense to add $prefix/lib to PATH when it comes to DLL search paths?

madebr: Yes, I'm looking into that atm.

I tried that first yesterday, but couldn't get it to work with [`msys2/setup-msys2@v2`](https://github.com/msys2/setup-msys2).
But I'm not using it anymore, so it's fine.

mikkelfj: But that didn't answer why tests a stored in the bin directory.

madebr: Another fine approach would be to use the `ENVIRONMENT` property on targets.
See https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html

This can be done very easily with the postponed test patches (=doing diff analysis).

madebr: > But that didn't answer why tests a stored in the bin directory.

It just avoided setting PATH.

mikkelfj: As I recall, all executables were built somewhere in the build tree, and only flatcc had a specific output target to bin.
Now it seems that all executables default to bin?

madebr: Yes, for mingw on WIndows.
I'm testing whether this can be reverted.

madebr: By the way, building these inside the source tree causes problems when configure cmake in 2 folders, as ci does for cross building.
Both might build `libflatcc.a` and `libflatccrt.a` in the `lib` subfolder.

Imho it's better to build these in a build tree subfolder.

mikkelfj: The logical thing would be to define a separate bin, lib location for cross target. They definitely should not be the same location.

madebr: That's a mess.

You can also have 2 native cmake build trees in parallel with different configuration paramenters.
Those would then conflict.
Keeping the binaries in the build tree is the logical thing.
You have the install target to export everything.

madebr:
I rebased on top of master.

@mikkelfj 
To avoid bitrot, can you please take a(nother) look at this?

mikkelfj:
Is it ready? I thought you had abandoned it.

madebr:
> Is it ready? I thought you had abandoned it.

I think so, I stopped pushing changes because I ran out of ideas 😄

mikkelfj:
OK, I'll have a look at some point. But I'll need to make a release before merging. Looking forward to CI fixes.

madebr:
I messed up my latest force push, but was able to recover using  c723a5e.

mikkelfj:
I finally got some time to work on flatcc. I'm going to close this PR because we have several build updates going on including PR #258. You are still welcome to comment on that PR and introduce your ideas, but I can't have two major PR on the build system. The plan is make a release soon, before major updates to the build system.

PR #177, opened 2021-03-26
Fix two typos in README

dariusarnold:
Hello, I noticed these two types while reading the docs and thought I could help out by fixing them.


mikkelfj:
Thanks, that is helpful.

PR #180, opened 2021-04-28
hexdump: constify

yamt:


mikkelfj:
Thanks, but shouldn't the `descr` argument also be const?

PR #186, opened 2021-05-21
Fix end_loc in json parser context

olivier-matz-6wind:
(Extracted from PR #185)

After parsing a table or struct, ctx->end_loc is set to the current
value of buf, which is expected to be the end of the json data.

However the ctx->end_loc value is incorrect, because buf is not updated
after the call to the parser callback function.

Fix this in both flatcc_json_parser_table_as_root() and
flatcc_json_parser_struct_as_root().


mikkelfj:
thanks

PR #187, opened 2021-05-21
Fix per-table file identifiers and extensions

olivier-matz-6wind:
(Extracted from PR #185)

These 2 commits change how per-table file identifiers and extensions are generated:
- the file identifiers should not use `flatbuffers_identifier` because it prevents from including several schema definitions, use a direct string instead
- the per-table file extensions are not generated, let's add them similarly to file identifiers

changes in v2:
- fix documentation (README.md)

### DIFF
--- a/config/config.h
+++ b/config/config.h
@@ -337,8 +337,9 @@
 #define FLATCC_DEFAULT_BIN_SCHEMA_EXT ".bfbs"
 #endif
 
+/* Schema file extensions do not carry a dot by convention, do the same here. */
 #ifndef FLATCC_DEFAULT_BIN_EXT
-#define FLATCC_DEFAULT_BIN_EXT ".bin"
+#define FLATCC_DEFAULT_BIN_EXT "bin"
 #endif
 
 #ifndef FLATCC_DEFAULT_DEP_EXT
--- a/README.md
+++ b/README.md
@@ -1280,8 +1280,8 @@ and so does other `_as_root` methods.
 
 The `file_extension` is handled in a similar manner:
 
-    #ifndef MyGame_Example_Monster_extension
-    #define MyGame_Example_Monster_extension flatbuffers_extension
+    #ifndef MyGame_Example_Monster_file_extension
+    #define MyGame_Example_Monster_file_extension "mon"
     #endif
 
 ### Type Identifiers
--- a/README.md
+++ b/README.md
@@ -1263,16 +1263,13 @@ defined as the null identifier.
 
 The generated code defines the identifiers for a given table:
 
-    #ifndef MyGame_Example_Monster_identifier
-    #define MyGame_Example_Monster_identifier flatbuffers_identifier
+    #ifndef MyGame_Example_Monster_file_identifier
+    #define MyGame_Example_Monster_file_identifier "MONS"
     #endif
-
-The `flatbuffers_identifier` is the schema specific `file_identifier`
-and is undefined and redefined for each generated `_reader.h` file.
 
 The user can now override the identifier for a given type, for example:
 
-    #define MyGame_Example_Vec3_identifier "VEC3"
+    #define MyGame_Example_Vec3_file_identifier "VEC3"
     #include "monster_test_builder.h"
 
     ...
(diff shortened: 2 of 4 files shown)


mikkelfj:
There is something strange with the dot prefix in the file extension. "mon" is not listed as ".mon" in monster_schema.fbs

In the log of the test build file_extension is found with ".cgen_test" in one place, and "cgen_test" without dot in some other place.

mikkelfj:
A reference output is printed without dot:
https://github.com/dvidelabs/flatcc/blob/2eaf0fa3a18e5fcab7fc66f04accb677119f2e27/test/cgen_test/cgen_test.c#L110

I wonder both why the test does not fail against the reference, and also why the dot is placed in the generated output.

mikkelfj:
The dot is generated here, in existing code:

https://github.com/dvidelabs/flatcc/blob/2eaf0fa3a18e5fcab7fc66f04accb677119f2e27/src/compiler/codegen_c_reader.c#L1012

I'm not sure that is correct to have a dot prefix, but it would be a breaking change to remove it.

olivier-matz-6wind:
> The dot is generated here, in existing code:
> 
> https://github.com/dvidelabs/flatcc/blob/2eaf0fa3a18e5fcab7fc66f04accb677119f2e27/src/compiler/codegen_c_reader.c#L1012
> 
> I'm not sure that is correct to have a dot prefix, but it would be a breaking change to remove it.

Yes, you are right, I'll add the dot in next version.

Note: in case we use the default bin (out->opts->default_bin_ext), the dot is already present, so no change needed for this code branch.

mikkelfj:
I think it would be reasonable to make a breaking change here and remove the dot from the output.

Also note that the prerecorded output uses "file_extension" where the generated output uses flatbuffers_extension. I guess there is no diff executing in the text.

mikkelfj:
> Note: in case we use the default bin (out->opts->default_bin_ext), the dot is already present, so no change needed for this code branch.

Yes the default is defined with a dot, but the question is whether it should be.

olivier-matz-6wind:
Without the dot, it looks more consistent with the schema.
I can try to remove the dot everywhere then (will do in another commit, before this one).

mikkelfj:
good

olivier-matz-6wind:
changes in v3:
- add a new commit to remove the dot from generated file extension (i.e. #define flatbuffers_extension)
- rework last commit to remove the dot too

I chose to keep FLATCC_DEFAULT_BIN_EXT and opts->default_bin_ext with the dot, for consistency with other extension macros and variables. The dot is removed when printed, using (opts->default_bin_ext + 1).
Let me know if you are ok with this.

mikkelfj:
+1 is not good without checking for length and dot.

I would have suggested to change the config file, but as you say, other ext vars use a dot prefix. There isn't really a good answer here. I would prefer if the dot was added in the schema and the default var kept as is, but that is not convention.

I think the best option is to remove dot from config var and add a comment that schema file identifiers do not carry a dot by convention. That way users can also define their own ext if they want.

olivier-matz-6wind:
> +1 is not good without checking for length and dot.

Oh yes, that's right

> I would have suggested to change the config file, but as you say, other ext vars use a dot prefix. There isn't really a good answer here. I would prefer if the dot was added in the schema and the default var kept as is, but that is not convention.
> 
> I think the best option is to remove dot from config var and add a comment that schema file identifiers do not carry a dot by convention. That way users can also define their own ext if they want.

Ok, will do, thanks

olivier-matz-6wind:
changes in v4:
- remove dot from FLATCC_DEFAULT_BIN_EXT config macro to avoid the unsafe (opts->default_bin_ext + 1)

config/config.h
```diff
@@ -337,8 +337,9 @@
 #define FLATCC_DEFAULT_BIN_SCHEMA_EXT ".bfbs"
 #endif
 
+/* schema file extensions do not carry a dot by convention, do the same here */
```
mikkelfj, line 340: Detail, but other comments start with uppercase and end with '.'

olivier-matz-6wind: I'm fixing it in v5

olivier-matz-6wind:
changes in v5:
- Fix comment style

mikkelfj:
thanks

olivier-matz-6wind:
thanks for the quick and efficient review!

PR #193, opened 2021-06-07
fix flatcc return code on include failure

olivier-matz-6wind:
When the file being compiled includes another one that does not exist,
the generation fails but the return code is 0:

  user@host:$ echo 'include "unexistant_file.fbs";' > schema.fbs
  user@host:$ flatcc schema.fbs
  error reading included schema file: attributes.fbs
  user@host:$ echo $?
  0

In flatcc_parse_file(), the ret variable is initialized to -1 at the
beginning of the function, but modified to 0 when fb_parse() is called
with success. Subsequent "goto done" do not set back ret to -1, so the
return value is 0 when the file is not found.

Let's remove the use of ret in the fb_parse() test, and remove the
uneeded affectations of ret above.

Signed-off-by: Olivier Matz <olivier.matz@6wind.com>


mikkelfj:
thanks

PR #199, opened 2021-07-22
fix potential memory leaks in fileio.c and semantics.c

Adsun701:
This fixes potential memory leaks that were detected by a quick run of cppcheck. In `fileio.c`, `memcpy()` will only copy from `prefix` to `path` if `prefix_len` is greater than 0, so copying from a potentially null pointer is avoided. In `semantics.c`, `index.len` is set accordingly based on `index.type` so that the former variable will always be initialized.

### DIFF
--- a/src/compiler/fileio.c
+++ b/src/compiler/fileio.c
@@ -62,8 +62,10 @@ char *fb_create_join_path_n(const char *prefix, size_t prefix_len,
         return 0;
     }
     n = 0;
-    memcpy(path, prefix, prefix_len);
-    n += prefix_len;
+    if (prefix_len > 0) {
+        memcpy(path, prefix, prefix_len);
+        n += prefix_len;
+    }
     if (path_sep) {
         path[n++] = '/';
     }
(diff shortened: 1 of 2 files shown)


mikkelfj:
Copy operation: yes, that is technically undefined behaviour for null pointers regardless of length. Fine to fix though it hardly matters in praxis.

index: I don't recall what the len field is used for - but is not used with coerce. I'm not sure if it semantically makes sense to assing a length for numerics if the length field is not set elsewhere for numerics.

However, the index value ought to be zero initialized.

Adsun701:
@mikkelfj Fixed with index.len set to 0.

mikkelfj:
I just researched in parser.c. It seems that the v->len is only used with type information where scalar types have length 1 and fixed length arrays can have other lengths. For normal integer parsing, the length field is not used and most likely defaults to zero. I think the length field was added relatively late to handle fixed length arrays.

Thus, your latest change is fine, but I think leaving the code as it were is more correct if you initialize with `fb_value_t index = { 0 };`.

Adsun701:
@mikkelfj Fixed, thanks for the input.

mikkelfj:
Thanks

PR #202, opened 2021-07-30
Remove strict prototypes from clang build

danhorner:
The build was failing on test functions missing prototypes.

Sample: 
```
FAILED: test/json_test/CMakeFiles/json_test_uq.dir/test_json.c.o
/Library/Developer/CommandLineTools/usr/bin/cc  -Itest/json_test/generated -I../../include -DFLATCC_REFLECTION=1 -Wstrict-prototypes -Wsign-conversion -Wconversion -std=c11 -pedantic -Wall -Wextra -Werror -g -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk -DFLATCC_JSON_PARSE_ALLOW_UNQUOTED=1 -MD -MT test/json_test/CMakeFiles/json_test_uq.dir/test_json.c.o -MF test/json_test/CMakeFiles/json_test_uq.dir/test_json.c.o.d -o test/json_test/CMakeFiles/json_test_uq.dir/test_json.c.o -c ../../test/json_test/test_json.c
../../test/json_test/test_json.c:142:20: error: this old-style function definition is not preceded by a prototype [-Werror,-Wstrict-prototypes]
int edge_case_tests()
```

### DIFF
--- a/test/emit_test/emit_test.c
+++ b/test/emit_test/emit_test.c
@@ -30,7 +30,7 @@ int dbg_emitter(void *emit_context,
     return 0;
 }
 
-int debug_test()
+int debug_test(void)
 {
     flatcc_builder_t builder, *B;
     float x[10] = { 0 };
@@ -48,7 +48,7 @@ int debug_test()
  * this assumes a very simple schema:
  * "table { time: long; device: ubyte; samples: [float]; }"
  */
-int emit_test()
+int emit_test(void)
 {
     /*
      * Note that there is some apparently unnecessary padding after 0x01
--- a/test/json_test/test_json.c
+++ b/test/json_test/test_json.c
@@ -139,7 +139,7 @@ failed:
 #define TEST_ERROR_FLAGS(fparse, fprint, x, err) \
     ret |= test_json(scope, (x), 0, err, (fparse), (fprint), __LINE__);
 
-int edge_case_tests()
+int edge_case_tests(void)
 {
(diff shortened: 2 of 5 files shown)


mikkelfj:
Thanks for providing a fix for this.

However, the entire build should not be affected by a test case.
I think the actual problem is that

```
int edge_case_tests()
{
...
}
```

should have been

```
int edge_case_tests(void)
{
...
}
```

Can you please check that, and update the PR accordingly if it works.

mikkelfj:
More specifically here, as reported by the error message:

test/json_test/test_json.c:142:20

danhorner:
Sure. 
Btw, I notice the flag is commented out on the gcc build with a note to similar effect, so you might find more of these as tests are added by gcc users: 

https://github.com/dvidelabs/flatcc/blob/5bc7c3512109e5fd6ea231fe1d76f09e1ce7a24e/CMakeLists.txt#L244-L246

mikkelfj:
Thanks.

The thing is, it is useful to have very strict compiler settings as long as they are reasonable because it ensures maximum portability. Otherwise there would but many more issues like the one you just reported because everyone has their policies and compiler versions.

Clang is generally fair. GCC assumes that you only use GCC and adapt to their conventions to a point where you either break code on other compilers, or make the code very ugly. That is why strict settings has to go on GCC and we rely on Clang to ensure the code remains relatively clean.

PR #204, opened 2021-08-03
build: repair the build where CMake is not on the path

compnerd:
When cmake is not on the path, the build would previously fail as it
hardcoded the name for the cmake binary, expecting to find it on the
path.  Use `CMAKE_COMMAND` instead to reference the cmake binary that
was used to configure the target.


mikkelfj:
Thanks. Note that there is a PR with a new CMake build that has to wait until a new release. You might want to keep an eye on that.

compnerd:
Thanks; this was needed for me to be able to build as I don't have CMake in my path.  I think it might be nice to be able to build out of the box in such an environment.

Issue #210, opened 2021-09-03
Size prefixed buffers with large alignment (e.g. vectors of doubles) do not pass verifier


bjornharrtell:
I'm unable to find this case exercised in the monster_test suite and I'm unsure about what is the correct way to use the API.

I've tried these  two variants:

```c
int gen_flatgeobuf1(flatcc_builder_t *B)
{
    double xy[] = { 0, 1 };
    //double z[] = { 0 };

    FlatGeobuf_Geometry_ref_t geometry;

    flatcc_builder_reset(B);

    flatbuffers_double_vec_ref_t xyRef = flatbuffers_double_vec_create(B, xy, 2);
    //flatbuffers_double_vec_ref_t zRef = flatbuffers_double_vec_create(B, z, 1);

    ns(Geometry_start(B));
    ns(Geometry_xy_add(B, xyRef));
    //ns(Geometry_z_add(B, zRef));
    geometry = ns(Geometry_end(B));

    ns(Feature_start_as_root_with_size(B));
    ns(Feature_geometry_add(B, geometry));
    ns(Feature_end_as_root(B));
    return 0;
}

int gen_flatgeobuf2(flatcc_builder_t *B)
{
    FlatGeobuf_Geometry_ref_t geometry;

    flatcc_builder_reset(B);

    ns(Geometry_start(B));
    ns(Geometry_xy_start(B));
    ns(Geometry_xy_push_create(B, 1));
    ns(Geometry_xy_push_create(B, 2));
    ns(Geometry_xy_end(B));
    //ns(Geometry_z_start(B));
    //ns(Geometry_z_push_create(B, 1));
    //ns(Geometry_z_end(B));
    geometry = ns(Geometry_end(B));

    ns(Feature_start_as_root_with_size(B));
    ns(Feature_geometry_add(B, geometry));
    ns(Feature_end_as_root(B));
    return 0;
}
```

Commenting in the lines in any of the two methods results in verifier error "vector header out of range or unaligned".

I have the full test case added in this branch https://github.com/bjornharrtell/flatcc/tree/flatgeobuftest.

(the schema etc is from my project https://github.com/flatgeobuf/flatgeobuf)

mikkelfj:
First of all, make sure the buffer you verify is on an sufficiently aligned memory location - it likely is given your findings your previous issue.

Try to move content creation to after start_as_root, and try to check return codes, e.g. null refs.

[...]

PR #213, opened 2021-09-04
Minimal reproduction test case for misaligned vector of double

bjornharrtell:
I believe I have narrowed down the problem with #210 further. It appears to happen only for buffers with size.

I've made new doublevec_test with a very minimal schema and two test cases, one run without size prefix and one with.. the latter fails with "vector header out of range or unaligned".

ctest --verbose output:

```sh
20: doublevec table:
20: 00000000  04 00 00 00 f4 ff ff ff  04 00 00 00 00 00 00 00  |................|
20: 00000010  06 00 08 00 04 00                                 |......|
20: doublevec table with size:
20: 00000000  1a 00 00 00 08 00 00 00  00 00 00 00 f4 ff ff ff  |................|
20: 00000010  04 00 00 00 00 00 00 00  06 00 08 00 04 00        |..............|
20: running debug doublevec test
20: doublevec buffer failed to verify, got: vector header out of range or unaligned
```

The table definition is:

```fbs
table DoubleVec {
  a: [double];
}
```

And the gen code is creating as root and an empty vector for a as it does not seem to matter if I fill it or not with something.

If I create two variants *without* setting the a vector at all both verifies and looks like this:
```sh
1: doublevec table:
1: 00000000  04 00 00 00 fc ff ff ff  04 00 04 00              |............|
1: doublevec table with size:
1: 00000000  0c 00 00 00 04 00 00 00  fc ff ff ff 04 00 04 00  |................|
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -19,6 +19,7 @@ add_subdirectory(json_test)
 add_subdirectory(emit_test)
 add_subdirectory(load_test)
 add_subdirectory(optional_scalars_test)
+add_subdirectory(doublevec_test)
 # Reflection can break during development, so it is necessary
 # to disable until new reflection code generates cleanly.
 if (FLATCC_REFLECTION)
--- a/test/doublevec_test/CMakeLists.txt
+++ b/test/doublevec_test/CMakeLists.txt
@@ -1,0 +1,20 @@
+include(CTest)
+
+set(INC_DIR "${PROJECT_SOURCE_DIR}/include")
+set(GEN_DIR "${CMAKE_CURRENT_BINARY_DIR}/generated")
+set(FBS_DIR "${PROJECT_SOURCE_DIR}/test/doublevec_test")
+
+include_directories("${GEN_DIR}" "${INC_DIR}")
+
+add_custom_target(gen_doublevec_test ALL) 
+add_custom_command (
+    TARGET gen_doublevec_test
+    COMMAND cmake -E make_directory "${GEN_DIR}"
+    COMMAND flatcc_cli -a -o "${GEN_DIR}" "${FBS_DIR}/doublevec_test.fbs"
+    DEPENDS flatcc_cli "${FBS_DIR}/header.fbs"
+)
+add_executable(doublevec_test doublevec_test.c)
+add_dependencies(doublevec_test gen_doublevec_test)
+target_link_libraries(doublevec_test flatccrt)
+
+add_test(doublevec_test doublevec_test${CMAKE_EXECUTABLE_SUFFIX})
--- a/test/doublevec_test/doublevec_test.c
+++ b/test/doublevec_test/doublevec_test.c
@@ -1,0 +1,113 @@
+#include <stdio.h>
+
+#include "doublevec_test_builder.h"
+#include "doublevec_test_verifier.h"
+
+#include "flatcc/support/hexdump.h"
+#include "flatcc/support/elapsed.h"
+#include "../../config/config.h"
+
+#undef nsc
+#define nsc(x) FLATBUFFERS_WRAP_NAMESPACE(flatbuffers, x)
+
+int gen_doublevec(flatcc_builder_t *B)
+{
+    flatcc_builder_reset(B);
+
+    DoubleVec_start_as_root(B);
+    DoubleVec_a_create(B, 0, 0);
+    DoubleVec_end_as_root(B);
+
+    return 0;
+}
+
+int gen_doublevec_with_size(flatcc_builder_t *B)
+{
+    flatcc_builder_reset(B);
+
+    DoubleVec_start_as_root_with_size(B);
+    DoubleVec_a_create(B, 0, 0);
+    DoubleVec_end_as_root(B);
+
+    return 0;
+}
+
+int test_doublevec(flatcc_builder_t *B)
+{
+    void *buffer;
+    size_t size;
+    int ret;
+
+    gen_doublevec(B);
+
+    buffer = flatcc_builder_finalize_aligned_buffer(B, &size);
+    hexdump("doublevec table", buffer, size, stderr);
+
+    if ((ret = DoubleVec_verify_as_root(buffer, size))) {
+        printf("doublevec buffer failed to verify, got: %s\n", flatcc_verify_error_string(ret));
+        return -1;
+    }
+
+    flatcc_builder_aligned_free(buffer);
+    return ret;
+}
+
+int test_doublevec_with_size(flatcc_builder_t *B)
+{
+    void *buffer, *frame;
+    size_t size, size2, esize;
+    int ret;
+
+    gen_doublevec_with_size(B);
+
+    frame = flatcc_builder_finalize_aligned_buffer(B, &size);
+    hexdump("doublevec table with size", frame, size, stderr);
+
+    buffer = flatbuffers_read_size_prefix(frame, &size2);
+    esize = size - sizeof(flatbuffers_uoffset_t);
+    if (size2 != esize) {
+        printf("Size prefix has unexpected size, got %i, expected %i\n", (int)size2, (int)esize);
+        return -1;
+    }
+
+    if ((ret = DoubleVec_verify_as_root(buffer, size))) {
+        printf("doublevec buffer failed to verify, got: %s\n", flatcc_verify_error_string(ret));
+        return -1;
+    }
+
+    flatcc_builder_aligned_free(frame);
+    return ret;
+}
+
+
+int main(int argc, char *argv[])
+{
+    flatcc_builder_t builder, *B;
+
+    (void)argc;
+    (void)argv;
+
+    B = &builder;
+    flatcc_builder_init(B);
+
+#ifdef NDEBUG
+    printf("running optimized doublevec test\n");
+#else
+    printf("running debug doublevec test\n");
+#endif
+#if 1
+    if (test_doublevec(B)) {
+        printf("TEST FAILED\n");
+        return -1;
+    }
+#endif
+#if 1
+    if (test_doublevec_with_size(B)) {
+        printf("TEST FAILED\n");
+        return -1;
+    }
+#endif
+
+    flatcc_builder_clear(B);
+    return 0;
+}
--- a/test/doublevec_test/doublevec_test.fbs
+++ b/test/doublevec_test/doublevec_test.fbs
@@ -1,0 +1,5 @@
+table DoubleVec {
+  a: [double];
+}
+
+root_type DoubleVec;


mikkelfj:
Great reproduction case.

First note that it could be completely random that one buffer works and the other doesn't if for example that aligment argument given the the create vector call was 4 instead of the expected 8.

Based on you first dump:

In the buffer without size prefix, the vector size field starts at offsets 0x000c (12 dec) and contains 4 zero bytes.
In the buffer with size prefix, the vector size field starts at offset 0x0014 (20 dec) and contains 4 zero bytes.

Observations: this size field must be aligned to 4 bytes. Because the vector is a double, the byte that follows the size field must be aligned to 8 bytes.

In the buffer without size field, the following address is 0x0010 (16 dec) which is a multiple of 8.
In the buffer with size field the following address is 0x0018 (24 dec) which is a multple of 8.

The vectors are therefore correctly aligned in both cases, assuming I have made no mistake in the above.

In the buffer without size prefix there is no zero padding after the header. That is coincidental because it just happened to not be necessary. The table starts at offset 0x0004 which is stored as the root table offset in the first 4 bytes of the table header at offset 0x0000. The table starts with the vtable offset which is the negative value "f4 ff ff ff" or 0xfffffff4, or -12 dec, meaning the vtable appears 12  bytes later. That is just how FlatBuffers work. And indeed we find the 6 byte long vtable at that location.

In the buffer with size prefix, there are 4 bytes of zero padding at offset 0x0008 (8 dec). The table starts at offset 0x000c (12 dec) storing a negative -12 value to the vtable 12 bytes later where the 6 byte vtable is found. The header stores the buffer size at offset 0x0000 with the value 0x0000001a (26 dec) which is the length of the printed buffer excluding the 4 byte size prefix itself. (I do not recall if the size prefix should include itself or not). **The root table offset is found at offset 0x0004 (4 dec) and contains the location 0x0008 (8 dec) - this wrong since this is the zero padded location.**

It appears that the root table offset is wrong when a size prefixed buffer needs global zero padding after the header.

mikkelfj:
I made a mistake in my conclusion - see edit above.

To further confirm that the buffer is valid, try access a vector with actual data in the buffer with size prefix and confirm the pointer addresses and content when reading the data with ordinary reader interface. Make sure the buffer is aligned including the size prefix.

bjornharrtell:
Thanks for your analysis @mikkelfj. I will see if I can confirm this by verifying the buffers with the C++ implementation.

mikkelfj:
I don't think C++ verifies aligment.

bjornharrtell:
@mikkelfj ok :( But I went ahead and did it anyway and got strange results...

The "empty" doublevec table verifies, i.e this passes:

```c++
const unsigned char flatbuf2[] =
{
  0x04, 0x00, 0x00, 0x00, 0xfc, 0xff, 0xff, 0xff,  0x04, 0x00, 0x04, 0x00
};
flatbuffers::Verifier verifier2((uint8_t *) flatbuf2, sizeof(flatbuf2));
TEST_EQ(VerifyDoubleVecBuffer(verifier2), true);
```

The buffer with an empty double vector in it but still *without* size prefix does not pass:

```c++
const unsigned char flatbuf3[] = 
{
  0x04, 0x00, 0x00, 0x00, 0xf4, 0xff, 0xff, 0xff,  0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
  0x06, 0x00, 0x08, 0x00, 0x04, 0x00
};
flatbuffers::Verifier verifier3((uint8_t *) flatbuf3, sizeof(flatbuf3));
TEST_EQ(VerifyDoubleVecBuffer(verifier3), true);
```

bjornharrtell:
I also tried producing buffers with the same content using C++ to compare. I know it doesn't have to be the same binary representation for the same content and I don't know enough about flatbuffers to understand it all yet but nevertheless:

`04 00 00 00 FC FF FF FF 04 00 04 00` - flatcc, doublevec empty no size prefix 
`08 00 00 00 04 00 04 00 04 00 00 00` - cC++, doublevec empty no size prefix 

Both the above verifies i C++ verifier.

`04 00 00 00 F4 FF FF FF 04 00 00 00 00 00 00 00 06 00 08 00 04 00` - flatcc, doublevec with empty a vector no size prefix
`0C 00 00 00 00 00 06 00 08 00 04 00 06 00 00 00 04 00 00 00 00 00 00 00`  - C++, doublevec with empty a vector no size prefix

Only the second verifies in the C++ verifier and it's two bytes longer.. (but that might be the tail pad to 8 bytes alignment in C++?)

The

bjornharrtell:
@mikkelfj while I'm far from sure about my findings above and not sure exactly what they mean it doesn't seem to confirm that the problem we see here with flatcc is restricted to the verifier.

mikkelfj:
The flatbuf3 buffer in your example has a size prefix of the vector which is 4 (at byte offset 8 in the buffer). That is followed by 4 zero bytes. If this is a byte vector it would work, but it is a double vector it makes no sense. The size is not a multiple of 8.

bjornharrtell:
@mikkelfj flatbuf3 buffer was created with this flatcc code:

```c
DoubleVec_start_as_root(B);
DoubleVec_a_create(B, 0, 0);
DoubleVec_end_as_root(B);
```

Are you saying we are back to conclude that there is indeed a bug with double vectors in flatcc?

mikkelfj:
Correction: the size field is not 4. 4 is the offset to the vector. So it is correct. The following zero bytes is the length field. In flatbuf3 the alignment is correct because the size starts at offset 12 and the empty vector payload starts at offset 16 which is a multiple of 8.

In flatbuf2 the offset to the vector to be "04 00 04 00" which is wrong.  It should be "04 00 00 00" and it should be followed by 4 zeroes and then by a vtable. Someting is missing.

mikkelfj:
In your comment starting with "I also tried producing buffers", The flatcc ( C ) and flatc ( C++ ) buffer, the buffers have the same content, but the vtable is stored before the table in C++. The results in more padding required which explains the longer length. In both cases the size of the vector is 4, which again makes no sense for empty double vector

mikkelfj:
In your comment starting with "I also tried producing buffers", I don't understand the first dumps. The second dumps: The flatcc ( C ) and flatc ( C++ ) buffer, the buffers have the same content, but the vtable is stored before the table in C++. The results in more padding required which explains the longer length. In both cases the offset the vector is 4 and points to a size field with zeroes that are a followed by an offset that is a multiple of 8. So both look correct.

As to the first dumps, I'll need more time but it appears to have "04 00 04 00" similar to before.

mikkelfj:
I think the first dumps are garbage.

mikkelfj:
I don't have time trace this right now, but I doubt that you are listing the buffers correctly. Also, without proper format hex dumps as in your previous posts, it is very difficul to assess alignment.

bjornharrtell:
Hmm ok.. strange.. I tried to be careful with the dumps but I had to find something to do it in C++ that's why the format isn't exactly the same.

One interesting fact though is that functionally this all appears to be working if I stay in flatcc land (correct roundtrip write/read and comparison) and avoiding the verifier - but causes alignment and possibly other memory issues when subjected to sanitizers. So whatever problem there is, it is likely systemic.

mikkelfj:
https://github.com/dvidelabs/flatcc/blob/master/doc/binary-format.md

might help

Some of the dumps are missing vtables like 06 00 08 00 04 00. That, and the strange 04 00 04 00 sequence is why I doubt the dumps.

mzaks:
There are two things that surprise me with `0x04, 0x00, 0x00, 0x00, 0xfc, 0xff, 0xff, 0xff,  0x04, 0x00, 0x04, 0x00`

1. The offset is a negative number, maybe it changed but AFAIK offset are defined as `u32` which is a waste because it is caped by `2^31` so `i32` makes more sense, but it is not compliant with the spec.
2. The vTable is broken. First `0x04, 0x00` of the vTable determine the size of the vTable, which is correct in this case, but actually shouldn't be. A vTable normally contains of following values first 2 bytes (vTable size), second 2 bytes (table data size) and then followed by 2 bytes relative offsets from table offset to the property values. In the buffer at hand there are only vTable size and kind table data size, 4 being the size of the offset to vTable itself. But I don't see the the relative offset to the property. I guess it can be inferred that as the vTable is just 4 bytes long, the relative property offsets are `0`. But that's quite implicit. I would expect the `0x00` to be part of the vTable and the padding which would have to be added in case of the odd number fo properties.

That sad in the C++ version: `0x08 0x00 0x00 0x00 0x04 0x00 0x04 0x00 0x04 0x00 0x00 0x00` The vTable is also "cropped" but the vTable offset is a positive number as I would expect.

mikkelfj:
@mzaks The offsets to the vtables are soffset_t which is a signed 32-bit value. The value is subtracted from the location where the offset is stored. Since C++ generally stores vtables before the tables that use them, the soffsets are typically small positive numbers. FlatCC stores vtables clustered together last in the buffer, so it stores negative values at table start.

mzaks:
`04 00 00 00 F4 FF FF FF 04 00 00 00 00 00 00 00 06 00 08 00 04 00` - flatcc, doublevec 0 len a vector no size prefix
Makes sense to me even though there is again the case of a negative offset, the table itself is:
`F4 FF FF FF 04 00 00 00` which is vTable offset + vector offset. the central `00 00 00 00` is the size of the vector. Then it says that vTabel is of size 6, the size of the table data is 8 and the relative offset to the first property is 4. Which is all correct values.

`0C 00 00 00 00 00 06 00 08 00 04 00 06 00 00 00 04 00 00 00 00 00 00 00` - C++, doublevec 0 len a vector no size prefix
is as far as I can tell correct.

First four bytes `0C 00 00 00` is offset to table data. The table data is `06 00 00 00 04 00 00 00`. First value 6 is the offset to vTable, which points us to `06 00 08 00 04` the second value 4 is the relative pointer to the vector which contains only of `00 00 00 00 00` because the size is 0. `00 00` between offset to table and vTable data are padding.

mikkelfj:
@mzaks 
> The vTable is broken. First 0x04, 0x00 of the vTable determine the size of the vTable, which is correct in this case,

That is not the vtable. That part of the buffer is missing. The 04 00 04 00 is still a strange value.

mzaks:
Ah, @mikkelfj I think you are correct, sorry. It is `i32` because of the vTable reuse feature. I forgot about it.

mikkelfj:
Your next analysis is correct from a cursory look. That is why I earlier said the second set of buffers appear correct, but the first do not make any sense.

mzaks:
> Your next analysis is correct from a cursory look. That is why I earlier said the second set of buffers appear correct, but the first do not make any sense.

it makes sense if `04 00 04 00` is a vTable which does not provide the relative property offsets ;).

mikkelfj:
> it makes sense if 04 00 04 00 is a vTable which does not provide the relative property offsets ;).

Yes it appears you are right.

mikkelfj:
@bjornharrtell that means the buffer dumps are probably correct. In that case they should also verify since the null tables cannot be misaligned as they are not present.

bjornharrtell:
Well, I have double checked and I cannot get `04 00 00 00 F4 FF FF FF 04 00 00 00 00 00 00 00 06 00 08 00 04 00` - flatcc, doublevec 0 len a vector no size prefix to verify with the C++ verifier (as done in https://github.com/dvidelabs/flatcc/pull/213#issuecomment-913147207).

And then we have the size prefixed variants from the initial description:

doublevec table with size (with empty vector):
```sh
20: doublevec table with size:
20: 00000000  1a 00 00 00 08 00 00 00  00 00 00 00 f4 ff ff ff  |................|
20: 00000010  04 00 00 00 00 00 00 00  06 00 08 00 04 00        |..............|
```

doublevec table with size (without vector):
```sh
1: doublevec table with size:
1: 00000000  0c 00 00 00 04 00 00 00  fc ff ff ff 04 00 04 00  |................|
```

The latter verifies with the C++ verifier as follows:
```c++
const unsigned char flatbuf5[] =
{
  0x0c, 0x00, 0x00, 0x00, 0x04, 0x00, 0x00, 0x00,  0xfc, 0xff, 0xff, 0xff, 0x04, 0x00, 0x04, 0x00
};
print_buf("flatbuf4", flatbuf5, sizeof(flatbuf5));
flatbuffers::Verifier verifier5((uint8_t *) flatbuf5, sizeof(flatbuf5));
TEST_EQ(VerifySizePrefixedDoubleVecBuffer(verifier5), true);
```

The former fails to verify:
```c++
const unsigned char flatbuf4[] =
  {
    0x1a, 0x00, 0x00, 0x00, 0x08, 0x00, 0x00, 0x00,  0x00, 0x00, 0x00, 0x00, 0xf4, 0xff, 0xff, 0xff,
    0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,  0x06, 0x00, 0x08, 0x00, 0x04, 0x00
  };
  print_buf("flatbuf4", flatbuf4, sizeof(flatbuf4));
  flatbuffers::Verifier verifier4((uint8_t *) flatbuf4, sizeof(flatbuf4));
  TEST_EQ(VerifySizePrefixedDoubleVecBuffer(verifier4), true);
```

mikkelfj:
I'd like to know more about why the C++ verifier fails.

As to C++ verifier failing on "The buffer with an empty double vector in it but still without size prefix does not pass:"

```
0x04, 0x00, 0x00, 0x00, 0xf4, 0xff, 0xff, 0xff,  0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
  0x06, 0x00, 0x08, 0x00, 0x04, 0x00
 ```

I really cannot see what would be wrong with that buffer except that the verifier might be looking for an optional file identifier. Without more information it is hard to tell.

As for those with size prefixes, perhaps we should focus on the one without first, but still, if you can get an error message, assertion or breakpoint clue ...

**Comment by mikkelfj** [comment]
Ohh

If C++ verifies alignment, then you cannot verify a buffer in a const char array. It will not be aligned. You need a uint64_t array or use alignas(8) attribute. The same goes for flatcc but since that would be a different process, it might get lucky and find aligned data.

**Comment by bjornharrtell** [comment]
Unfortunately the C++ verifier does not seem to have any means of error reporting unless I'm missing something and I need to refresh skills in gdb to provide any other leads.

About the const char array, if I memcpy it to a freshly allocated buffer would that suffice?

**Comment by mikkelfj** [comment]
yes

**Comment by bjornharrtell** [comment]
Ok so back to the non size prefixed buffer with zero sized vector in it.. still do not verify (not sure I got the malloc / memcpy part right for alignment issues.. not sure how I otherwise can allocate aligned buffer in C++):

```c++
const unsigned char flatbuf3chars[] =
{
  0x04, 0x00, 0x00, 0x00, 0xf4, 0xff, 0xff, 0xff,  0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
  0x06, 0x00, 0x08, 0x00, 0x04, 0x00
};
void *flatbuf3 = (uint8_t *) malloc(sizeof(flatbuf3chars));
memcpy(flatbuf3, flatbuf3chars, sizeof(flatbuf3chars));
flatbuffers::Verifier verifier3((uint8_t *) flatbuf3, sizeof(flatbuf3chars));
TEST_EQ(VerifyDoubleVecBuffer(verifier3), true);
```

mikkelfj:
memcpy looks legit.

bjornharrtell:
Strange.. it's indicated here https://github.com/google/flatbuffers/issues/4774 that the C++ verifier should assert in debug builds but I only get false return false.

bjornharrtell:
May have been able to debug myself to where the C++ verifier concludes that `04 00 00 00 F4 FF FF FF 04 00 00 00 00 00 00 00 06 00 08 00 04 00` (22 bytes) does not pass. AFAIK it's because it is checking "message size" at:

https://github.com/google/flatbuffers/blob/550d2f9048ce2a9d572da14a36043cf3f5b87bdd/include/flatbuffers/flatbuffers.h#L2535

And that function returns 0 because at:
https://github.com/google/flatbuffers/blob/550d2f9048ce2a9d572da14a36043cf3f5b87bdd/include/flatbuffers/flatbuffers.h#L2594 size is 24 and size_ is 22.

The C++ generated buffer with the same content is 24 bytes so it seems to be there is some assumption on length/padding/alignment here that do not hold with the flatcc made buffer...

mikkelfj:
OK, that is also technically correct if C++ expects a tailpadded buffer. You can just try again:

`asize = (size + 7) & ~(size_t)7` and allocate as much.

bjornharrtell:
@mikkelfj you are probably right, padding with two zero bytes makes it pass C++ validator:

```c++
uint8_t *flatbuf3 = hexstr_to_char(
"04 00 00 00 f4 ff ff ff  04 00 00 00 00 00 00 00"
"06 00 08 00 04 00 00 00", 24);
print_buf("flatbuf3", flatbuf3, 24);
flatbuffers::Verifier verifier3(flatbuf3, 24);
VerifyDoubleVecBuffer(verifier3);
TEST_EQ(VerifyDoubleVecBuffer(verifier3), true);
```

mikkelfj:
Can you please also check the with size cases?

I'll debug the flatcc verifier with size when I get around to it. Meanwhile, I suggest confirming that the buffer both verifies with C++ and produce valid data when read by flatcc, and ignore any verifier errors produced for with size buffers.

bjornharrtell:
@mikkelfj yes will do asap

mikkelfj:
BTW: please this PR more or less as is since I use it for testing.

bjornharrtell:
The "flatcc, doublevec empty with size prefix" passes C++ verification:

```c++
//  flatcc, doublevec empty with size prefix
uint8_t *flatbuf5 = hexstr_to_char(
  "0c 00 00 00 04 00 00 00  fc ff ff ff 04 00 04 00", 16);
print_buf("flatcc, doublevec empty with size prefix", flatbuf5, 16);
flatbuffers::Verifier verifier5(flatbuf5, 16);
TEST_EQ(VerifySizePrefixedDoubleVecBuffer(verifier5), true);
```

The "flatcc, doublevec 0 len a vector with size prefix" does not pass C++ verification:

```c++
// flatcc, doublevec 0 len a vector with size prefix
uint8_t *flatbuf4 = hexstr_to_char(
  "1a 00 00 00 08 00 00 00  00 00 00 00 f4 ff ff ff"
  "04 00 00 00 00 00 00 00  06 00 08 00 04 00", 30);
print_buf("flatcc, doublevec 0 len a vector with size prefix", flatbuf4, 30);
flatbuffers::Verifier verifier4(flatbuf4, 30);
//TEST_EQ(VerifySizePrefixedDoubleVecBuffer(verifier4), true); // fails verification
```

Attempt to manually add tail padding "flatcc, doublevec 0 len a vector with size prefix + manual two byte padding" still does not pass C++ verification:

```c++
// flatcc, doublevec 0 len a vector with size prefix + manual two byte padding
uint8_t *flatbuf44 = hexstr_to_char(
  "1a 00 00 00 08 00 00 00  00 00 00 00 f4 ff ff ff"
  "04 00 00 00 00 00 00 00  06 00 08 00 04 00 00 00", 32);
print_buf("flatcc, doublevec 0 len a vector with size prefix + manual two byte padding", flatbuf44, 32);
flatbuffers::Verifier verifier44(flatbuf44, 32);
//TEST_EQ(VerifySizePrefixedDoubleVecBuffer(verifier44), true); // fails verification
```

I've tried to debug the C++ verifier in the above case but I haven't been able to pinpoint yet.

bjornharrtell:
Debugging session was correct, adjusting the size prefix (which make sense now.. ) makes is pass:

```c++
// flatcc, doublevec 0 len a vector with size prefix + manual two byte padding + adjusted size prefix
uint8_t *flatbuf44 = hexstr_to_char(
  "1c 00 00 00 08 00 00 00  00 00 00 00 f4 ff ff ff"
  "04 00 00 00 00 00 00 00  06 00 08 00 04 00 00 00", 32);
print_buf("flatcc, doublevec 0 len a vector with size prefix + manual two byte padding", flatbuf44, 32);
flatbuffers::Verifier verifier44(flatbuf44, 32);
TEST_EQ(VerifySizePrefixedDoubleVecBuffer(verifier44), true);
```

mikkelfj:
OK thanks, good to know.

bjornharrtell:
Just to be clear from what I hear the minimum alignment of a flatbuffer is 4 bytes or the size of the largest scalar stored and the full buffer size should always be divisible with the minimum alignment, so what you call tail padding is not optional it is mandatory in the flatbuffer specification (or reference implementation).

mikkelfj:
The min alignment is 4 because of the uoffset type. The actual alignment will typically be the largest scalar actually stored in the buffer, so either 4 or 8. However, the alignment can be larger if a struct or vector with larger alignment is stored in the buffer. Flatcc currently only supports specifying struct alignment beyond its natural content based alignment. The alignment is always a power of 2. I have not heard of it being divisible by min alignment, but that kind of follows from the power of 2.

bjornharrtell:
> The min alignment is 4 because of the uoffset type. The actual alignment will typically be the largest scalar actually stored in the buffer, so either 4 or 8. However, the alignment can be larger if a struct or vector with larger alignment is stored in the buffer. Flatcc currently only supports specifying struct alignment beyond its natural content based alignment. The alignment is always a power of 2. I have not heard of it being divisible by min alignment, but that kind of follows from the power of 2.

Agreed. I was just trying to make it clear that flatcc makes buffers that to not adhere to the alignment requirement i.e a doublevec table with an empty vector of doubles without size prefix is produced to a 22 byte buffer but should be 8 byte aligned. This is already deduced by the above experiments I just wanted to get the conclusion down and confirmed as a bug.

mikkelfj:
I wouldn't call it a bug. But FlatBuffer were never very precisely specified and it was an active design choice. C++ tail alignment cannot be exact unless it always pads up to the maximum possible alignment. For example some buffers might store a 256 byte aligned struct for use with graphics card, while other buffers of same type might not. The alignment does not affect the current buffer, but the following buffer. Hence it does not really make sense to tail pad based on current buffer content unless you alway opt for max. That sounded a bit excessive so flatcc chose to not do that.

In hindsight flatcc should probably have padded to 4 or 8 bytes just to ensure that buffers can be stacked meaningfully, but since that is easy to do manually, there wasn't really a need for it. Also, size prefixed buffers did not exist early on.

In reality one ought to prepad buffers up to the alignment requirement of the buffer being added to a stack of buffers. That will ensure proper alignment. So the more I think about it, the more I think it is best to just leave it alone. If you want to verify with C++, just align the buffer manually.

Flatcc gives you the alignment requirement of the current buffer just before or after finalize so prepadding a stacked buffer is easy.

mikkelfj:
What is a bug is that the flatcc verifier fails a valid size prefixed buffer. That needs to be fixed.

mikkelfj:
Also, tail aligment needs a section in documentation to explain what I just stated in the above.

mikkelfj:
Of course there is still the issue with adjusting the size prefix, but that cannot really be helped, except with a helper function to adjust an existing buffer based on the next buffers alignment requirement.

aardappel:
It is pretty simple, this is a difference with how it is implemented in C++, which is the "gold standard" as far as how FlatBuffers should work, given any ambiguities in the text specification. C is doing something different from pretty much all other language implementations here, and fails the C++ verifier. So regardless of whether theoretically this would make sense to be allowed, this should be fixed. The eco-system does not benefit from one implementation doing things incompatibly from the others.

mikkelfj:
@aardappel I tend to agree, but can you give the exact requirements the C++ implements?

aardappel:
`minalign` is defined as the size of the largest scalar serialized, or the largest pointer to a scalar that may need to be created (in the case or an empty vector), or similarly, the largest data item with a `force_align` attribute.

All such data items are self-aligned to both the start and the end of the buffer, thus the size of the buffer is a multiple of `minalign`.

mikkelfj:
Thanks, that is essentially the alignment reported by the flatcc builder near completion.

mikkelfj:
I am starting to look into this again.

To summarize:
Tail padding discussed above should be fixed, but it is beside the point of the issue being discussed, although it affects comparisons with C++ verifier.

The issue is simple in principle:
At least some parts of the verifier tests for alignment relative to buffer start, notably the vector header alignment wrt. the element size even if the vector is empty.
The verifier has not been specialized for _with_size (the optional header size prefix), and it should not need to because the data inside must still be aligned.
This fails because some checks are relative to buffer start rather than pointer address.
Why? Because if the buffer start is 8 byte aligned at the size prefix, it is verified 4 bytes later to skip the size prefix (this is added in the user code seen above). Now suddenly the offset to buffer start appears unaligned.

The fix is likely to add specific with_size variants of the verifier and feed it the original buffer start with size prefix.

mikkelfj:
@aardappel I have now added code to pad flatcc buffers to the largest object alignment. There was actually alignment code already, but it requires that the user calls start_buffer with a block alignment argument, but it is not set by default.

But I don't think this is quite right either. If a buffer sometimes contains large objects (e.g. double) and large aligned structs, and sometimes not, then buffers will have different alignments, and when stacking buffers with size prefix, the buffer following might have a larger object and be incorrectly aligned.

The solution is to find the theoretical maximum alignment in the schema compiler. The flatcc builder should already be prepared to take such an alignment argument, but it needs to be generated first. But this could also be wasteful if some buffers have rarely used large alignments.

aardappel:
I guess the crucial difference is that so far we have made no guarantee that if you put multiple FlatBuffers consecutively in memory that this guarantees their alignment, like structs would. We only guarantee alignment for a single FlatBuffer, storing multiple FlatBuffers is between you and your memory allocators, etc.

We could make this "work by default" by aligning a FlatBuffer not to the largest stored item, but the largest declared item. I would be against such a change, given how:

1. This causes all sorts of buffers to grow bigger unnecessarily. It is very common for large parts of a schema tree to be optional, and for people to include schemas of other teams/people etc. Imagine an optionally stored 16-byte aligned vector suddenly bloating all buffers even though it is hardly ever used.
2. Would be a ton of work to guarantee is actually adhered to across all our language implementations.
3. Would not be backwards compatible when new large-aligned items are added to the schema.
4. Would not be backwards compatible with all existing written FlatBuffer data.

So, an implementation should default to "largest stored item". That said, if an implementation has users that would like to "stack" FlatBuffers, an optional "largest declared item" mode could be allowed, since more padding is entirely compatible with the existing FlatBuffers design. But it should not be the default. If you made it the default, then users of said implementation would run into trouble the moment they accept data made by another language. The user would need to be very aware that the alignment they are requesting is not guaranteed across FlatBuffers.

@dbaileychess to see if he agrees with my reasoning.

mikkelfj:
I agree to that. This is also largely how flatcc has worked, except it did not tailpad until recently. The user could always either self-align before stacking, or passing a block-align argument to the builder.

mikkelfj:
Just for the record: As far as I can tell from the flatcc code, older version should actually tailpad correctly, but ONLY if a block align argument was explicitly assigned other than the default 0 argument. Now the runtime largest block is taken into account when tailpadding. At any rate the largest runtime alignment has always been taking into account for prefix padding so the buffer starts correctly.

dbaileychess:
I agree with Wouter. Any sort of "meta" flatbuffer work (stacking multiple flatbuffers) should be outside the core library concern. We can provide helper accessors to get the largest declared type if needed.

mikkelfj:
The test case provided in this PR was useful in adding new _with_size verifiers. See end of #210 for details.

Note that previously tail pad was added which likely helps passing C++ verifier, but that was not the core issue.

PR #217, opened 2021-10-25
Newer CMake is going to drop <2.8.12

powderluv:
So update to 2.8.12.2 (in Ubuntu 14.04)

CentOS 7 has a way to install CMake3.

This avoids the warning in newer CMakes:

flatcc/CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,8 @@
 # Ubuntu 14.04 (Trusty)
-#cmake_minimum_required (VERSION 2.8.12.2)
+cmake_minimum_required (VERSION 2.8.12.2)
 # Centos 7
 #cmake_minimum_required (VERSION 2.8.11)
-cmake_minimum_required (VERSION 2.8)
+#cmake_minimum_required (VERSION 2.8)
 
 # Disable build of tests and samples. Due to custom build step
 # dependency on flatcc tool, some custom build configurations may


midokura-xavi92:
Relying on _ancient_ versions of CMake such as `2.8` has more serious implications. For example, CMake policy [`CMP0077`](https://cmake.org/cmake/help/latest/policy/CMP0077.html) was introduced in `3.13` so that higher-level `CMakeLists.txt` can adjust `option()`s programmatically using `set()`. Otherwise, the settings from the following example would be ignored by `flatcc`:

```cmake
cmake_minimum_required(VERSION 3.13.5)
project(test_project)
set(FLATCC_TEST OFF) # This will be ignored since flatcc currently uses cmake_minimum_required(VERSION 2.8)
add_subdirectory(flatcc)
```

Typically, CMake would then return the following warning message:
```
Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
```

If `flatcc` does not bump `cmake_minimum_required` to `3.13`, the only solution is to move all configurations as command line options e.g.:

```sh
$ cmake .. -D FLATCC_TEST=OFF
```

Which shifts responsibility to end users (who might be _or not_ maintainers of the higher-level `CMakeLists.txt`), potentially causing confusion and issues.

mikkelfj:
Historically, certain important Ubuntu versions did not trivially work with more recent CMake.
I'm all for moving forward now. Especially since flatcc has just had a new release.
As I mentioned, there is a rather large pending PR on upgrading CMake already.
However, I cannot invest a lot of time in this for the time being.
At the same time, the build system for flatcc needs to be migrated to GH Actions. Especially for *nix builds, but for consistency also Windows builds.
Anyone willing to take on this, speak up.

midokura-xavi92:
> As I mentioned, there is a rather large pending PR on upgrading CMake already.

I had a look at https://github.com/dvidelabs/flatcc/pull/171, but IMHO this PR introduces _too many_ several architectural changes that affect many aspects of this project i.e., not only build-related stuff, but also integration with Github Actions, cross-compilation and introduction of new APIs such as `flatcc_generate_sources`, among others.

I would rather suggest to split the effort in #171 into smaller, independent PRs that make it easier for others to review. Therefore, as a first step, I plan to create a PR that identifies obsolete and deprecated CMake constructs within the project and fixes them with modern alternatives.

mikkelfj:
I tend to agree, but I'd rather have GH Actions working first because the Travis build stopped working after they made a change. So there is no way to verify complex changes across several operating systems and compiler versions as it is.
Thanks for having a look.

mikkelfj:
@midokura-xavi92 don't let this stop you if you have something of interest. This is just the order of events I think makes the most sense.

midokura-xavi92:
@mikkelfj thank you very much for your support. I will then provide a few suggestions whenever possible.

mikkelfj:
This was merged into PR #250 so closing this one.
@midokura-xavi92 if you still have some input on CMake changes, the new GH Actions makes it easier to test changes.

midokura-xavi92:
Sorry for not bringing attention to this topic earlier - priorities were shifted to other projects, so no time could be dedicated to [this suggestion](https://github.com/dvidelabs/flatcc/pull/217#issuecomment-1145862231), unfortunately, and I do not know when I could take back on it.

PR #222, opened 2021-11-23
Fix #221 Bad string lenght cause segmentation fault.

### LINKED ISSUE #221: Bad binary format may cause segmentation fault when call function verify
Hi,

In binary format, a bad length string can cause segmentation fault.
With sample monster, change "Orc" string lenght from 0003 0000 to FFFF FFFF, the call of verify crash.

```
00000000  04 00 00 00 84 ff ff ff  00 00 80 3f 00 00 00 40  |...........?...@|
00000010  00 00 40 40 30 00 00 00  1c 00 00 00 0c 00 00 00  |..@@0...........|
00000020  2c 00 00 00 2c 01 00 01  02 00 00 00 2c 00 00 00  |,...,.......,...|
00000030  1c 00 00 00 0a 00 00 00  00 01 02 03 04 05 06 07  |................|
00000040  08 09 00 00 ff ff ff ff  4f 72 63 00 d4 ff ff ff  |........Orc.....|
00000050  14 00 00 00 05 00 00 00  e0 ff ff ff 10 00 00 00  |................|
00000060  03 00 00 00 03 00 00 00  41 78 65 00 05 00 00 00  |........Axe.....|
00000070  53 77 6f 72 64 00 00 00  08 00 0a 00 04 00 08 00  |Sword...........|
00000080  18 00 24 00 04 00 00 00  20 00 10 00 00 00 14 00  |..$..... .......|
00000090  22 00 18 00 23 00 1c 00                           |"...#...|

```

Geneo-5:
If n = 0xFFFFFFFF then  n + 1 = 0.
end - base (unsigned value) is always greater then 0.
Finally the function test then value (buf + base)[0xFFFFFFFF] whish
crash.

### DIFF
--- a/src/runtime/verifier.c
+++ b/src/runtime/verifier.c
@@ -265,7 +265,7 @@ static inline int verify_string(const void *buf, uoffset_t end, uoffset_t base, 
     base += offset;
     n = read_uoffset(buf, base);
     base += offset_size;
-    verify(end - base >= n + 1, flatcc_verify_error_string_out_of_range);
+    verify(end - base > n, flatcc_verify_error_string_out_of_range);
     verify(((uint8_t *)buf + base)[n] == 0, flatcc_verify_error_string_not_zero_terminated);
     return flatcc_verify_ok;
 }


mikkelfj:
I'm not sure that end - base is always greater than zero, but it should always be greater than zero due handle zero termination. Can you explain?
I think you fix is correct though.

Geneo-5:
The offset is type unsigned int 32.
Let n = 0xFFFFFFFF
For all **end** and **base** in [0, 0xFFFFFFFF], **end** - **base** >= n + 1 because n + 1 = 0
The last test check if the end of string is set to 0x00 but with this n, the test is outside the buffer memory (buffer + base + 0xFFFFFFFF)

mikkelfj:
I understand. But I do not understand your initial comment "end - base (unsigned value) is always greater then 0."
Do you mean that it cannot be be zero, or that it must not be zero? Because I think it can be zero even if the buffer is not valid then.

PR #228, opened 2022-03-29
disable pedantic for GCC >= 8

le91688:
disable -pedantic flag for GCC version > 8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,7 +179,7 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pedantic -Wall -Wextra")
     # Fix broken C++ alignas - either will do
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-    #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DPORTABLE_PATCH_CPLUSPLUS_STDALIGN") 
+    #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DPORTABLE_PATCH_CPLUSPLUS_STDALIGN")
     if (FLATCC_ALLOW_WERROR)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
     endif()
@@ -209,8 +209,10 @@ elseif (CMAKE_C_COMPILER_ID STREQUAL "GNU")
         endif()
     else()
         message(STATUS "Setting GNU C compiler options with c11 and Posix")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -pedantic -Wall -Wextra")
-        if (NOT (GCC_VERSION VERSION_LESS 8.0))
+        if (GCC_VERSION VERSION_LESS 8.0)
+            message(STATUS "Disabling GNU C pedantic for GCC < 8.0")
+            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wall -Wextra")
+        elseif (NOT (GCC_VERSION VERSION_LESS 8.0))
             # Disable some GCC checks:
             # (warnings exist since 8.0, but are more aggressive in 9.0)
             #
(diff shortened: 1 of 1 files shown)


mikkelfj:
No that is orthogonal.

You can probably make an if then else construct here:

https://github.com/dvidelabs/flatcc/blob/07ae7dca8118f9ab6d900a7d4797881cab708ba6/CMakeLists.txt#L211-L213

pjr-z:
> No that is orthogonal.
> 
> You can probably make an if then else construct here:
> 
> https://github.com/dvidelabs/flatcc/blob/07ae7dca8118f9ab6d900a7d4797881cab708ba6/CMakeLists.txt#L211-L213

I thought the issue was happening with clang?

le91688:
> No that is orthogonal.
> 
> You can probably make an if then else construct here:
> 
> https://github.com/dvidelabs/flatcc/blob/07ae7dca8118f9ab6d900a7d4797881cab708ba6/CMakeLists.txt#L211-L213

Ah my apologies, should have waited for your response in the issue thread!

Im happy to address the GCC version flag disable, but what i was hoping to accomplish was a generalized way to disable pedantic ( in my particular case, im struggling getting flatcc runtime to build for linux-arm / linux-mips with Clang )

mikkelfj:
We could a add a FLATCC_ flag to disable pedantic. The problem is that every time there is a new GCC release, one or more users report some issue that is is a complete non-issue, and it needs to be investigated and GCC needs to be silenced. And with a flag, the user then needs to be told to add that flag to the build environment. GCC pendantic is just broken in my view.

mikkelfj:
>  ( in my particular case, im struggling getting flatcc runtime to build for linux-arm / linux-mips with Clang )

Are you saying this a pedantic error raised by Clang?
In that case we need to fix the source of the problem in pattributes because we need clang pedantic since it usually find problematic areas. This is also why we can afford to remove on GCC.

le91688:
> We could a add a FLATCC_ flag to disable pedantic. The problem is that every time there is a new GCC release, one or more users report some issue that is is a complete non-issue, and it needs to be investigated and GCC needs to be silenced. And with a flag, the user then needs to be told to add that flag to the build environment. GCC pendantic is just broken in my view.

Understood. How about a FLATCC_NOPEDANTIC flag, and we set that flag for GCC versions over a specified target?

le91688:
> > ( in my particular case, im struggling getting flatcc runtime to build for linux-arm / linux-mips with Clang )
> 
> Are you saying this a pedantic error raised by Clang? In that case we need to fix the source of the problem in pattributes because we need clang pedantic since it usually find problematic areas. This is also why we can afford to remove on GCC.

Yes, thats correct, this is a clang error im experiencing with fall through.

mikkelfj:
OK, lets take this dicussion back to the issue.

le91688:
@mikkelfj  updated as per our discussion

mikkelfj (changes requested):

mikkelfj:
Can you test this with a few different GCC versions? The Travis build has stopped working so I don't have any insights on CI.

le91688:
> Can you test this with a few different GCC versions? The Travis build has stopped working so I don't have any insights on CI.

sure, will test 8, 10, 11

le91688:
@mikkelfj  , tested (via `rikorose/gcc-cmake` docker images) gcc 5 , gcc 8, and gcc 11 

command used:
`cmake -DFLATCC_INSTALL=on -DCMAKE_INSTALL_PREFIX="/test/" -DFLATCC_RTONLY=on -DCMAKE_BUILD_TYPE=Release && make install`

mikkelfj:
Thanks. It would be better if you could run the full tests, or at least a build without RTONLY, see scripts/test.sh, but at least it shows the compilers don't go belly up.

le91688:
> Thanks. It would be better if you could run the full tests, or at least a build without RTONLY, see scripts/test.sh, but at least it shows the compilers don't go belly up.

sure i can do that quick

mikkelfj (approved):

le91688:
Tested with gcc 5 and gcc 11, running `scripts/test.sh` successfully

mikkelfj:
Thanks, looks good!

mikkelfj:
@pjr-z I missed your comment on clang, could have saved some effort, thanks for the heads up tho :)

PR #237, opened 2022-07-06
Add undefined behaviour sanitizer and fix some undefined behaviour

r-barnes:
This fixes some instances of undefined behaviour. Specifically:
```
[1/85] Running utility command for gen_monster_test_prefix
/z/flatcc/src/compiler/parser.c:1276:19: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /z/flatcc/src/compiler/parser.c:1276:19 in 

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /z/flatcc/external/hash/cmetrohash.h:62:22 in 
/z/flatcc/src/compiler/codegen_c.c:168:17: runtime error: null pointer passed as argument 2, which is declared to never be null

/z/flatcc/src/runtime/builder.c:208:13: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /z/flatcc/src/runtime/builder.c:208:13 in 
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,10 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang")
     if (FLATCC_IGNORE_CONST_COND)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-tautological-constant-out-of-range-compare")
     endif()
+    if (CMAKE_BUILD_TYPE MATCHES Debug)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined")
+    endif()
 
     # To get assembly output
     # set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -save-temps")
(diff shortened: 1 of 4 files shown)


dontlaugh:
Aren't these flags only available on certain versions of clang? Do you need to guard against that in CMake?

mikkelfj:
Ah sorry, I have a backlog to attend to. Not sure about the flags, but they should only be set on compilers that support them, obviously, including a version check. In this case there is probably no reason to add it to anything but clang since clang runs checks in CI and the should do.

mikkelfj:
@r-barnes thansk for following up. I'm just superficially revisiting for now - I should have more time in a few weeks. It is not sufficient that flags work, they need to work with specific versions. FlatCC supports very old compilers too. You can see this in the root CMakeList.txt.

mikkelfj:
@r-barnes Lets keep this until I get more time to look at it. If there is UB this should be fixed or at least understood, flags or not. After that we can discuss if there should be changes to build, but maybe not.

mikkelfj:
@r-barnes Thanks for the contribution. I'm sorry I have not had time to follow up before now.
The flags are indeed reasonable for clang debug and not really needed for other compilers.

I have made a separate commit 0e925e1 to address this and I am therefore closing this PR.

Some fixes to remove warnings were already done, and some fixes were done differently (adding 0 to null ptr. was handled via cast in macro instead of a runtime check in builder).

Issue #231, opened 2022-05-10
flatcc cli compiles with clang but not with gcc on Linux


MartinKlang:
Current master branch doesn't build with gcc without `-Wno-sign-conversion`:
```
$ CC=gcc cmake ..
$ make
...
[ 16%] Building C object src/compiler/CMakeFiles/flatcc.dir/semantics.c.o
/home/mars/devel/evp-device-agent/evp_agent/sdkenc/flatcc/src/compiler/semantics.c: In function ‘analyze_struct’:
/home/mars/devel/evp-device-agent/evp_agent/sdkenc/flatcc/src/compiler/semantics.c:620:25: error: unsigned conversion from ‘int’ to ‘uint16_t’ {aka ‘short unsigned int’} changes the value of ‘-2’ [-Werror=sign-conversion]
  620 |     ct->symbol.flags &= ~fb_circular_open;
      |                         ^
cc1: all warnings being treated as errors
...
```

It does build with clang:
```
$ CC=clang cmake ..
$ make
...
[ 16%] Building C object src/compiler/CMakeFiles/flatcc.dir/semantics.c.o
[ 17%] Building C object src/compiler/CMakeFiles/flatcc.dir/coerce.c.o
[ 18%] Building C object src/compiler/CMakeFiles/flatcc.dir/flatcc.c.o
...
[100%] Built target bfbs2json
```
I'm on Ubuntu 22.04, gcc 11.2.0, clang 14.0.0, cmake 3.22.1, GNU Make 4.3

mikkelfj:
Also, even if it is unsigned, it would still be a truncation to uint16_t on most platforms, so a cast is probably needed regardless.

MartinKlang:
It's just that one line, if I cast to uint16_t then it builds correctly.

MartinKlang:
With the cast, the tests pass
```
CXX=g++ CC=gcc scripts/test.sh
...
-- The C compiler identification is GNU 11.2.0
-- The CXX compiler identification is GNU 11.2.0
...
100% tests passed, 0 tests failed out of 22

Total Test time (real) =   0.46 sec
TEST PASSED
```

MartinKlang:
```
diff --git a/src/compiler/semantics.c b/src/compiler/semantics.c
index 96d4b94..09b98d3 100644
--- a/src/compiler/semantics.c
+++ b/src/compiler/semantics.c
[...]

PR #248, opened 2022-10-18
Use an unsigned type for public interface flags

bjosv:
When using the provided API to set multiple JSON-printer flags via `flatcc_json_printer_set_flags()`
some compilers and tools will give errors due to a bitwise operation on a signed integer type.

The provided flags are an enum and in C enums are signed (C99 6.4.4.3p2).
Since C doesn't mandate how to represent a negative signed value (one's complement/two's complement/..)
there is no portable way to do bit operations on signed values.

When a user wants to set multiple flags using bitwise OR with the provided flags this gives errors in the user code indicated by tools like clang-tidy. See the checker:
https://releases.llvm.org/10.0.0/tools/clasng/tools/extra/docs/clang-tidy/checks/hicpp-signed-bitwise.html

This PR proposes the use of unsigned int flags instead of an enum,
which avoids having problems in the usercode when using this library.

An example project that triggers the issue:
https://github.com/bjosv/flatcc-experiments/tree/main/issues/enums
see [comment](https://github.com/bjosv/flatcc-experiments/blob/c7d317dae52046b7dcf56a0872db75431f578e09/issues/enums/main.c#L27
).

A C++ reference regarding the use of unsigned types for bit manipulations:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-unsigned

### DIFF
--- a/include/flatcc/flatcc_builder.h
+++ b/include/flatcc/flatcc_builder.h
@@ -710,10 +710,13 @@ static inline void flatcc_builder_refmap_reset(flatcc_builder_t *B)
 }
 
 
-enum flatcc_builder_buffer_flags {
-    flatcc_builder_is_nested = 1,
-    flatcc_builder_with_size = 2,
-};
+typedef uint16_t flatcc_builder_buffer_flags_t;
+static const flatcc_builder_buffer_flags_t flatcc_builder_is_nested = 1;
+static const flatcc_builder_buffer_flags_t flatcc_builder_with_size = 2;
+
+/* The flag size in the API needs to match the internal size. */
+static_assert(sizeof(flatcc_builder_buffer_flags_t) ==
+              sizeof(((flatcc_builder_t *)0)->buffer_flags), "flag size mismatch");
 
 /**
  * An alternative to start buffer, start struct/table ... end buffer.
@@ -776,7 +779,7 @@ enum flatcc_builder_buffer_flags {
 flatcc_builder_ref_t flatcc_builder_create_buffer(flatcc_builder_t *B,
         const char identifier[FLATBUFFERS_IDENTIFIER_SIZE],
         uint16_t block_align,
-        flatcc_builder_ref_t ref, uint16_t align, int flags);
+        flatcc_builder_ref_t ref, uint16_t align, flatcc_builder_buffer_flags_t flags);
 
 /**
  * Creates a struct within the current buffer without using any
@@ -867,7 +870,7 @@ flatcc_builder_ref_t flatcc_builder_end_struct(flatcc_builder_t *B);
  */
 int flatcc_builder_start_buffer(flatcc_builder_t *B,
         const char identifier[FLATBUFFERS_IDENTIFIER_SIZE],
-        uint16_t block_align, int flags);
+        uint16_t block_align, flatcc_builder_buffer_flags_t flags);
 
 /**
  * The root object should be a struct or a table to conform to the
@@ -923,7 +926,7 @@ flatcc_builder_ref_t flatcc_builder_end_buffer(flatcc_builder_t *B, flatcc_build
  */
 flatcc_builder_ref_t flatcc_builder_embed_buffer(flatcc_builder_t *B,
         uint16_t block_align,
-        const void *data, size_t size, uint16_t align, int flags);
+        const void *data, size_t size, uint16_t align, flatcc_builder_buffer_flags_t flags);
 
 /**
  * Applies to the innermost open buffer. The identifier may be null or
--- a/include/flatcc/flatcc_json_parser.h
+++ b/include/flatcc/flatcc_json_parser.h
@@ -22,13 +22,12 @@ extern "C" {
 #define PDIAGNOSTIC_IGNORE_UNUSED
 #include "flatcc/portable/pdiagnostic_push.h"
 
-enum flatcc_json_parser_flags {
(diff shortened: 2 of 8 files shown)


mikkelfj:
From memory I think this issue has been up before. If I remember correctly enums are so widely used in flatcc that is not practical to change to unsigned constants. Fixing the JSON printer would in that case only be a band aid.

There should be an issue discussing this somewhere.

mikkelfj:
You can see an example below:

The solution is to cast wherever logical operations are applied is shown in the comment. Note that this isn't cosmetic - it is potentially undefined behavior or something along this lines.

https://github.com/dvidelabs/flatcc/issues/231#issuecomment-1122449195
quote
```diff --git a/src/compiler/semantics.c b/src/compiler/semantics.c
index 96d4b94..09b98d3 100644
--- a/src/compiler/semantics.c
+++ b/src/compiler/semantics.c
@@ -617,7 +617,7 @@ static int analyze_struct(fb_parser_t *P, fb_compound_type_t *ct)
     }
 
     ct->symbol.flags |= fb_circular_closed;
-    ct->symbol.flags &= ~fb_circular_open;
+    ct->symbol.flags &= (uint16_t)~fb_circular_open;
     ct->order = P->schema.ordered_structs;
     P->schema.ordered_structs = ct;
     return ret;```

**Comment by mikkelfj** [comment]
Note: I would not mind a PR that fixes this, but it is presumably a LOT of work to cover all the enums.

**Comment by bjosv** [comment]
>  If I remember correctly enums are so widely used in flatcc that it is not practical to change to unsigned constants. 

I have seen the use of enums in many places, but I guess the issue is only given when it's used as a flag by the user, which I haven't seen that many cases of.
Possibly when using `flatcc_builder_buffer_flags` which we haven't used, but maybe that is another candidate that also should be changed in this case..

**Comment by mikkelfj** [comment]
Well, a limited fix to user facing flags would be fine. After all, internal use can, and have, been handled with casts.

Of course it might affect users as the interface would break slightly, but that is OK.

We just need to make sure we find all the cases. Both JSON printer and parser at least, and probably several places in the builder / emitter / allocator. I.e. inspection of all header files in include/flatcc.

**Comment by mikkelfj** [comment]
As to your PR, I think the proposed changes are fine as far as they go. I am just wondering if we should use a fixed size type instead of unsigned int since unsigned int could technically be too small for some values on some (hypothetical) compilers. We could also consider adding a typedef for the flag type. The generated code with flatbuffer schema enums does something similar.

**Comment by bjosv** [comment]
Sounds good, I can add commits with changes for `flatcc_builder_buffer_flag` and `flatcc_json_parser_flags` to this PR as well, unless you want to have it separate or do it yourself. There might be more that these two, but I'll take a deep dive..

For the type I'm open for suggestions I you prefer any, or have some previous pattern.

**Comment by mikkelfj** [comment]
Please go ahead. I think one PR is fine - if it becomes huge, maybe more would be needed. Note that I generally squash the PR into master.

**Comment by mikkelfj** [comment]
If you have 3 clean commits for printer, parser and buffer I would use those over squashing.

**Comment by mikkelfj** [comment]
Wrt type
I'm thinking something like
`typedef uint32_t flatcc_json_parser_flags_t;`
but may also be uint16_t depending on the type used internally. If values are close to limit the type may have to be larger for future extension.

**Comment by bjosv** [comment]
I could only find the printer, parser and buffer enums which now is replaced.

The buffer flag has a size dependency to the internals but should be covered by a static assert (which seems to be covered the portability layer), but the placement of it can maybe discussed..

**Comment by mikkelfj** [comment]
Yes, I looked through the code, can say if it is all correct, test must confirm that, but it looks solid.
I also noticed the static assert. I'm not sure if it is necessary and if the null pointer cast is sufficiently portable.

Note: if you can the run the test with clang and gcc on Linux it would be good since only appveyor windows CI/CD is currently working. I'll check MacOS.

**Comment by bjosv** [comment]
I have run fine with:
```
11333  2022-10-18 18:30:33  ./scripts/test.sh\n
11334  2022-10-18 18:30:50  CXX=clang++ CC=clang ./scripts/test.sh\n
```
on Linux (Ubuntu 22.04.1 LTS) with:
- Ubuntu clang version 14.0.0-1ubuntu1
- g++ (Ubuntu 11.2.0-19ubuntu1) 11.2.0

bjosv:
I'm not that familiar with appveyor, is it better than githubs CI? Might have to look into that.

Regarding the assert then I was suddenly thinking if it should be moved away from the "public" headers that users includes, since its an internal thing. Maybe it can trigger warnings in usercode by some static tool..
But its nice to have something since I add the cast, which would hide a single handed changes in the future.

mikkelfj:
Well, we can leave until someone complains ...

Appveyor is for historical reasons - the only way to test on Windows. Travis was used fo MacOS & Linux Clang & GCC, but they changed their API/URL and I never got to update.

The plan is to move to Github now that that is available. There is a vacant position for that ...

bjosv:
Sure, sound good regarding the assert.
I ran clang-tidy on it with big bunch of checks and current seems fine.

Aha, yes, a lot of projects had to move from travis. Github supports windows and mac as well.. so if I'm bored by other things the I know where to look :)

bjosv:
Ah ok, windows is not my strong side, have only setup simple jobs..

Thanks for your effort with this project, and being so quick to respond!

mikkelfj:
Thanks. Anyway, a GH build for just Linux would  also be valuable.

PR #249, opened 2022-10-24
Fix "unused local variable" and "constant in condition" compilation warnings Fix compilation warnings 

kfbehar:
Under strict compilation modes, build reports "unused local variable" and "constant in condition" warnings.

This fix is a contribution from Microsoft.


mikkelfj:
Could you please provide some more background information, such as the compiler error, compiler version and platform. Also, some notes on why changes are necessary in each commit.
I understand the ret variable is probably unused, but it is hard to review without more information.

The __ASSERT_REASON__ macro - where is it defined. It is unlikely from flatcc runtime library and thus likely not portable. What is it trying to fix?

kfbehar:
> Could you please provide some more background information, such as the compiler error, compiler version and platform. Also, some notes on why changes are necessary in each commit. I understand the ret variable is probably unused, but it is hard to review without more information.
> 
> The `__ASSERT_REASON__` macro - where is it defined. It is unlikely from flatcc runtime library and thus likely not portable. What is it trying to fix?

yea, sorry about that, pull request was sent prematurely.
provided some more information.

mikkelfj:
>yea, sorry about that, pull request was sent prematurely.
provided some more information.

Thats fine, take your time. Just keep in mind that the project is used by all sorts of different platforms.

PR #250, opened 2022-10-24
Add Github CI

bjosv:
Add Github CI and weekly builds to run tests with a range of different compilers and compiler versions on Ubuntu, Mac and Windows.

This PR includes:
- Updated minimum required CMake version from 2.8 to 2.8.12.2
- Fixes to build with GCC on Mac.
- Fixes to build with Clang 3.9, 4.0, 5.0
- Additional fixes in test for GCC 32-bit


bjosv:
Is this inline with what we need in the project?
Currently the build would fail due to some compiler versions don't compile or the test fails, but I will look into that.
Some build logs:
Ubuntu: https://github.com/Nordix/flatcc/actions/runs/3315747970
Mac: https://github.com/Nordix/flatcc/actions/runs/3315747963

mikkelfj:
Haven't reviewed yet, but this sounds awesome.
GCC on Mac is somewhat pointless since it is just Clang pretending. But since you ran into issues, maybe it is not pointless after all. Strange with Clang errors? But could be regressions due to CI downtime.

mikkelfj:
MAcOS GCC, note line 311 in log: ` 1/22 Test  #1: monster_test_cpp .................Subprocess aborted***Exception: `. All subsequent errors with buffer too small is likely because something goes wrong when writing the buffer.

mikkelfj:
We also the need the banner for the build in README.

I am wondering if we should limit builds to certain branches - I think it is probably not constructive in this context although I strongly prefer it in production because it might otherwise trigger unwanted deployments etc.. However, the old Travis did how to build types: one with just the latest few compilers and one that did the whole thing, but it took a long time and I only used it before making a release to check against regressions. It also depends on GH speed.

I looked at the Ubuntu GH script. I strongly prefer Ninja for development, but I think Make is better for cross platform validation. I think the Travis build used Make. Ideally at least a few configurations would do both, but that might be overkill.

So this is more thoughts than requirements.

mikkelfj:
It is the ci-more branch where I add more test on Linux and macOS. There is something similar for Windows on Appveyor, but I think it stranded a bit because Appveyor changed their environment so I just chose the most important middle of the road builds there.

mikkelfj:
Also, if you haven't already. C++ tests are important. There is only a single test that can be disabled, but I got a lot of bug reports from C++ users until someone contributed that test. This avoids small regressions that breaks C++ that I wouldn't normally check for. Note that C++ test just means a regular C test included in C++ file and compiled with g++ or clang in C++ mode. I'm note sure C++ pass all compiler versions, but at least the recent ones and possible some older that currently works.

mikkelfj:
As to ci-more - if we keep that concept, I definitely would want only one version the GH script with branch detection within the script rather than merge controlled differences as it is today - that is very fragile.

mikkelfj:
Maybe the large build can be scheduled weekly or something?

bjosv:
Sure weekly works as well, this would be good if external analysis tools like https://scan.coverity.com/ is added later.
Coverity is pretty nice and gives pretty good warnings, but it has the caveat to only allow few runs per day/week for free. I have a weekly run of it using Github Actions in another project. Might be something for the future.

bjosv:
Some updates to have a CI run of basic targets and a weekly run with a wider range.

- Checked that all except gcc-4.4 runs the C++ test.
- Added usage of make for some jobs but maybe I should add some more in the weekly runs.
- Added a job for intels icc compiler handled in the CMakeLists
- Fixed the issues found regarding gcc on mac, and in older Clang on Linux

Maybe there could be some more runs of windows in weekly..

Examples:
https://github.com/Nordix/flatcc/actions/runs/3469147023
https://github.com/Nordix/flatcc/actions/runs/3469147024

mikkelfj:
This looks great. If we are going to switch from AppVeyor, we definitely need more 32 and 64 bit Windows builds weekly, down to at least MSVC 2013. Currently we go back to MSVC 2010, but I'm ready to drop that.

As long as you keep the commits clean I think it is fine as is, but otherwise source code fixes should preferably be in a separate PR.

Wish list:
I'm not sure if we did have 32 bit Clang or GCC builds, but I think a 32-bit build is in place to ensure integer types behave as they should.
Eventually ARM builds would also be more relevant.

mikkelfj:
BTW: I assume you noticed the annotations in yourr weekly build example:
https://github.com/Nordix/flatcc/actions/runs/3469147024

bjosv:
32-bit is a good idea, I'm not sure the best way to get the -m32 flag in there though, maybe one should give cmake a toolchain file or something.

I looked a bit at ARM but most are using own runners, but a cross-compilation seems to work which can be added.

I'll see if I can fix some of the annotation as well, the used install-clang-helper gives most warnings.

mikkelfj:
ARM is probably most interesting on an ARM OS where tests can also run.
But plenty of users have used ARM based systems with not issues so far, or at least the issues have been elsewhere with stdlib version etc.

mikkelfj:
I looked at the build logs from ICC in your weekly example. There are warnings that do not surface in the build overview (annotations), for example:

```
/home/runner/work/flatcc/flatcc/src/runtime/builder.c(1918): warning #188: enumerated type mixed with another type
      return B->frame ? frame(type) : flatcc_builder_empty;
             ^

/home/runner/work/flatcc/flatcc/src/runtime/builder.c(1926): warning #188: enumerated type mixed with another type
      return B->frame[level - B->level].type;
             ^
```

Also, it may be difficult to continue building with ICC, given the following from same build:

```
cc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.
```

mikkelfj:
Can you incorporate #217 (update CMake version) in this PR? Could easily add after, but it would be good to know that it works on all builds.

bjosv:
Current example runs:
CI: https://github.com/Nordix/flatcc/actions/runs/3491255367
Weekly https://github.com/Nordix/flatcc/actions/runs/3491255369

README.md
```diff
@@ -1,5 +1,6 @@
-OS-X & Ubuntu: [![Build Status](https://travis-ci.org/dvidelabs/flatcc.svg?branch=master)](https://travis-ci.org/dvidelabs/flatcc)
+Ubuntu, macOS and Windows: [![Build Status](https://github.com/dvidelabs/flatcc/actions/workflows/ci.yml/badge.svg)](https://github.com/dvidelabs/flatcc/actions/workflows/ci.yml)
 Windows: [![Windows Build Status](https://ci.appveyor.com/api/projects/status/github/dvidelabs/flatcc?branch=master&svg=true)](https://ci.appveyor.com/project/dvidelabs/flatcc)
+[![Build Status](https://github.com/dvidelabs/flatcc/actions/workflows/weekly.yml/badge.svg)](https://github.com/dvidelabs/flatcc/actions/workflows/weekly.yml)
```
mikkelfj, line 3: Missing a description for the weekly build I think.

mikkelfj:
I assume that you are still looking at 32-bit build, but otherwise - there is no test running for gcc 32-bit, only a check for the ./bin/flatcc file if I read it correctly. I also made on comment on a missing description for the weekly build status in README.

mikkelfj:
The Intel build runs both ICC ICX in the same build. I wonder if there is any good reason for that and how that affects reporting.

mikkelfj:
I also wonder if we should entirely skip the build for 2.8, or upgrade it to 2.8.12?

mikkelfj:
I note that ICX has zero warnings in build log, and ICC has an almost endless amount. Since Intel is completely switching over to ICX (clang based) I guess we can safely ignore those warnings.

README.md
bjosv: Here is an example how the image looks:
image text
I skipped a description since the text Weekly is included in the image, but maybe there is a need for a description to get some distance to the windows badge.

bjosv:
> I assume that you are still looking at 32-bit build, but otherwise - there is no test running for gcc 32-bit

I got problems with 2 tests in the test runs ([log](https://github.com/Nordix/flatcc/actions/runs/3495095895/jobs/5851488819)) with I'll look at.

> The Intel build runs both ICC ICX in the same build.

I put them in the same job to avoid downloading the compiler and duplication of the prepare step, but it can be split in two if you think that is better. One drawback with current single job this is that if icc fails then there is no attempt to build with icx. If the job is duplicated then they run in parallel indifferent of each other.

> I also wonder if we should entirely skip the build for 2.8, or upgrade it to 2.8.12?

CMake 2.8.12.2 is used in both cases, but when I set 2.8.12.2 for the actions-setup-cmake it fails to find it weirdly enough. Using 2.8.12 works when tested, so I'll change it to that. Will file an issue regarding this problem to the actions-setup-cmake project.

README.md
mikkelfj: Well, I cannot see how it looks yet so I am just raising a potential concern, but I think it makes sense that all three builds have some text since otherwise it might see as if it is related to the Windows build, I guess. You could call extended build or something.

mikkelfj:
My main concern with ICC if it is unsupported in the future and we want to stick with ICX, it is unclear what is necessary in the earlier build step in terms of downloads. And then of course if one fails the other is blind. So I think it is best to split since it is only weekly. It might also be possible to trigger a single sub build when tracking problems, not sure.

bjosv:
Building with 32bit with clang seems fine, so there is something with gcc that triggers problems with a float compare in an assert (emit_test) and something else in monster_test.

mikkelfj:
I'm looking at it right now, none the wiser:
https://github.com/dvidelabs/flatcc/blob/821cfa648e4d8ba50d411b7f1620ed7bea60d616/test/emit_test/emit_test.c#L115

It suggests we should probably have both a clang and a gcc 32 bit test?
Mind you, MSVC has been running 32-bit tests successfully for a very long time on Appveyor.

bjosv:
Adding -ffast-math to GCC seems to make it work due to changed precision, described a bit in [link](https://nullprogram.com/blog/2018/05/01/).
But I'm not an expert in this area..

mikkelfj:
The other test that fails is:
https://github.com/dvidelabs/flatcc/blob/821cfa648e4d8ba50d411b7f1620ed7bea60d616/test/monster_test/monster_test.c#L374-L378

bjosv:
Yes, in monster_test when printing out the values with `%a`
in GCC it becomes:
if (`-0x1.99999ap+1` != `-0x1.999999999999ap+1`) {
but from clang
if (`-0x1.99999ap+1` != `-0x1.99999ap+1`) {

mikkelfj:
point 4. in the this post:
https://stackoverflow.com/a/10335601

Seems like potentially a double rounding issue from high to low precision?

mikkelfj:
> if (-0x1.99999ap+1 != -0x1.999999999999ap+1) {

When you write 3.2f and not 3.2d, you mean mean 3.2f, how hard can it be?

mikkelfj:
Anyway, printf is not a good test - not sure about %a, but as I recall it only takes double as argument, so value has to be cast to double implicitly before printing.

mikkelfj:
I think we need to fix the tests somehow, just not sure how. -ffast-math seams a bit like a moving target.

bjosv:
Yes, printf if probably not the best test, as you say, there is probably some casting done..

mikkelfj:
We could add parse_float_is_equal and parse_double_is_equal to
https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pparsefp.h

mikkelfj:
Proposed addition to https://github.com/dvidelabs/flatcc/blob/master/include/flatcc/portable/pparsefp.h

```C
/* Work around gcc double precision conversion of float values. */
#ifdef GCC_... tbd

static inline int parse_float_is_equal(float x1, float x2)
{
    union { uint32_t u32; float f32; } v1;
    union { uint32_t u32; float f32; } v2;
    v1.f32 = x1;
    v2.f32 = x2;
    return v1.u32 == v2.u32;
}

#else

static inline int parse_float_is_equal(float x1, float x2)
{
    return x1 == x2;
}

#endif

static inline int parse_double_is_equal(double x1, double x2)
{
    return x1 == x2;
}
```

bjosv:
Nice, I'll try adding that.

mikkelfj:
Note I changed gcc to GCC in comment.

mikkelfj:
If that doesn't work we might have to add some EPSILON to each input value before casting to u32 to force lower bits to flip.

mikkelfj:
Some more background here:
https://bitbashing.io/comparing-floats.html
But I or we can improve if we get the tests running and plugging in the compare function.

mikkelfj:
Also a some discussion here:
https://stackoverflow.com/questions/5989191/compare-two-floats

mikkelfj:
I added parse_float_compare and parse_float_is_equal to pparsefp.h file on master.
Currently it is always taking slow path on 32-bit, i.e. no GCC specialization.

mikkelfj:
I pushed fixes to emit_test and monster_test on master. Also improved on the initial pparsefp.h impl.
Comparison now drops least significant bit in tests via parse_float_is_equal.

bjosv:
Ah great! Had to look at something else for some time, will rebase it in.

mikkelfj:
Just quick feedback on stuff noticed:
32 bit build steps are named something like "run build", while the others are "run build and test".
32-bit builds (and windows) does apparently not test both debug and release while scripts/test.sh used in most build variants should run both.

bjosv:
Updates in weekly:
- split Intel icc and icx jobs (via a matrix)
- CMake version specified as 2.8.12 (results in 2.8.12.2 which cant be set explicitly, some bug)
- 32bit builds on both GCC and Clang (current default version)

Added a fix for some more changes needed in monster_test.
Added description to readme.

Logs from latest run:
Weekly: https://github.com/Nordix/flatcc/actions/runs/3497788685
CI: https://github.com/Nordix/flatcc/actions/runs/3497788688

mikkelfj:
And maybe the min required build should be pushed back to runs on Ubuntu 20.04. I think 18.04 might fail. Then again, who knows how long Github supports that build ...

mikkelfj:
Oh, we are very close to have this going now ...
Thanks for all the hard work so far.

bjosv:
Just as a note, I cant (re)build 32bit and 64bit builds at the same time even using separate build-directories.
It complains about the libflatccrt.a lib is incompatible, which is put in a common place /lib.
I was just wondering about the debug/release variants could have some problem due to this..or maybe this has be thought thru already.

mikkelfj:
No issue, debug uses libflatccrt_d.a

mikkelfj:
I know someone once comlained about binary naming. I don't recall exactly but I think it was wrt. installing 32 and 64 bit binaries into official distribution dirs on some Linux. I kind of rejected on grounds that current setup follows CMake convention, and a distribution can change naming/flags as needed.

bjosv:
To avoid manually setup the build folders for 32bit builds it would be nice to use `scripts/initbuild.sh` to get the debug/release directories setup automagically, but I need to get the `-m32` flags in there somehow..
Maybe we could add a new build.cfg.xxx flag that has something like
`FLATCC_BUILD_FLAGS="-DCMAKE_C_FLAGS=-m32 -DCMAKE_CXX_FLAGS=-m32"` for initbuild.sh
Not sure how it fits..

mikkelfj:
I was thinking the same thing. It's just such a long time, so I don't remember anything. If you can add something to scripts/config, it would be great.

Note: the build kind of assumes the platform default, which by now is implicitly 64-bits. But it might not be for all users. This needs to be handled if making a specific 32-bit variant. This is also one reason why naming the builds to some specific beyond d_d for ebug is a bit tricky.

mikkelfj:
We can also add an optional secondary flag to initbuild.sh for arch, but maybe its better to just keep it in the build cfg name.

bjosv:
This works at least, and the -m32 flag should even work ok on an x86-32 according to [link](https://github.com/randombit/botan/issues/1438).
WDYT?

Weekly: https://github.com/Nordix/flatcc/actions/runs/3498433
CI: https://github.com/Nordix/flatcc/actions/runs/3498433065

mikkelfj:
Me thinx this looks good. As long as it is understood that make-32bit is a Unix style build since flags likely won't work with Windows. With that in mind, README build section might need a comment.

mikkelfj:
Any reason why Windows doesn't build in weekly?

bjosv:
> 

Only because of my lack of knowledge of building in windows :) 
I see now (if I understand correctly) that we only run VS 2015 in appveyor, so I guess there could be a range of different Visual studio versions used in weekly somehow.

bjosv:
I see that [setujp-msbuild](https://github.com/marketplace/actions/setup-msbuild) used in CI can take a VS version, I can look into that

mikkelfj:
Yeah, the Appveyor required different build images for different versions after 64-bit became dominant. The ci-more should do something differently, but it isn't really ideal there anymore.

I was just thining having some Windows build such that we don't have to both run ci and weekly to be reasonably confident that the things largely works.

I might add more windows builds over time, but it's been a while since I looked into it.

mikkelfj:
I suggest we just copy the Windows build to weekly for now, update README if relevant, then merge to master. After that we can polish the Windows build if possible.

mikkelfj:
But I won't stop you if you have something better for Windows.

bjosv:
Ok, seemed that it was not possible to get any version [anyway](https://github.com/microsoft/setup-msbuild/issues/95#issuecomment-1307256906), but
`For windows-latest or windows-2022 that will be Visual Studio 2022. And for windows-2019 that will be Visual Studio 2019`

Can copy the job we have, and run on both images to get two VS versions at least.

mikkelfj:
Also, you might want to update your initial PR comment since it isn't entirely up to date, yet rather specific.

mikkelfj:
I have tested running builds on much older MSVC versions on GH actions, but it is a bit involved and some time ago, so I don't want to go into that for now. Also, it was not with CMake. But I have some material I can dig into.

bjosv:
Perfect!
I'll try to tying up the loose ends and finalize this later, or during the weekend, it's time for a beer now :)

mikkelfj:
Great, and thanks again for the huge effort.

bjosv:
Unfortunately the weekend flew by.., but here is the changes:
- Added Windows to weekly
- Added Mac Clang to weekly
- Some notes in README about the new build configuration available via `initbuild.sh`
- Simplified the inital PR comment

mikkelfj:
This looks good. Some quick observations:

Windows has warnings in pparsefp, likely due to changes I made - these warnings are also hard to find in GH actions - appveyor is much better at destilling these.

ci build seems to run Ubuntu 20, while this is relevant for some old test, we probably want to bump to 22 or so, if possible, for latest compiler tests?

icc compiler warnings appear to be concentrated on a single const expression warning, so perhaps we can find and disable that warning?

GCC 4.4 has the following warning:
/home/runner/work/flatcc/flatcc/src/compiler/semantics.c:1549: warning: missing initializer
/home/runner/work/flatcc/flatcc/src/compiler/semantics.c:1549: warning: (near initialization for ‘index.<anonymous>.s.len’)
Maybe not too important, but worth investigating.

Overall, I think this is good enough to merge now, what do you think?

mikkelfj:
You can see the Appveyor warnings here:
https://ci.appveyor.com/project/dvidelabs/flatcc/build/job/d6bt1ss79irvrt86/messages

Do you think GH Actions has some similar filter?

mikkelfj:
Also, I think weekly.yml is currently set to build on push and pull requests.

bjosv:
Ah, the warnings-tab in appveyor looks usable. I haven't seen any similar on GH Actions yet though..
I'll update to ubuntu22 in CI before we merge.

The `on:` in the files describes when to run the workflow, and in CI its on `push` and `pull_requests`.
In weekly its on `workflow_dispatch` which enables manual runs of the workflow from the gui, and then the `schedule`.
I had to add `push` and `pull_request` temporary when testing, but to me it looks like I managed to remove them before pushing to the PR branch. The examples are taken from the branch with the temporary change.

mikkelfj:
I'm just thinking: I don't want weekly to run on PR because it takes too long for feedback, and trigger a lot of builds with frequent pushes, on other hand, it currently only takes 6 minutes to build, and I probably cannot trigger builds in PR branches since it is not my repo, or?

Most of the time I would assume things are good if ci.yml is good, but there might be cases where portability is more of a concern.

bjosv:
You are probably right, I've never attempted to run a workflow manually on a PR before, so it might be problematic. Not sure..cant find any examples describing it either.
This would for sure be nice to do when a PR is believed to affect a lot..

mikkelfj:
Any idea how to add the ci.yml build to PR checks? I had this for Travis and Appveyor, and recently just Appveyor, but I am a bit at a loss how to enable this with GH actions?

mikkelfj:
Or is it already enabled for future PRs? I don't see it on existing PRs.

bjosv:
So an update by committer or admin like [this](https://github.blog/changelog/2022-02-03-more-ways-to-keep-your-pull-request-branch-up-to-date/#update-your-pull-request-branch-by-rebasing) _should_ make it run.

mikkelfj:
OK, thanks, we'll see how it goes.

PR #252, opened 2022-11-08
Remove use of the `fallthrough` macro

### LINKED ISSUE #247: The internal define `fallthrough` leaks to usercode
When a source file includes both `flatcc` and  `Boost` the following problem appears:
```
staging/include/boost/config/compiler/clang.hpp:314:26: error: missing ')' after '('
#if !__has_cpp_attribute(fallthrough) || __cplusplus < 201406L
                         ^~~~~~~~~~~
staging/include/flatcc/portable/pattributes.h:74:22: note: expanded from macro 'fallthrough'
# define fallthrough pattribute(fallthrough)
                     ^~~~~~~~~~~~~~~~~~~~~~~
staging/include/flatcc/portable/pattributes.h:69:23: note: expanded from macro 'pattribute'
#define pattribute(x) pattribute_##x
                      ^~~~~~~~~~~~~~
<scratch space>:358:1: note: expanded from here
pattribute_fallthrough
^~~~~~~~~~~~~~~~~~~~~~
staging/include/flatcc/portable/pattributes.h:63:46: note: expanded from macro 'pattribute_fallthrough'
# define pattribute_fallthrough __attribute__((__fallthrough__))
                                             ^
staging/include/boost/config/compiler/clang.hpp:314:25: note: to match this '('
#if !__has_cpp_attribute(fallthrough) || __cplusplus < 201406L
                        ^
1 error generated.
```

Is there a suggestion of a remedy to a problems like this?

I am new to flatcc's code, but there seem to be a use of both `fallthrough` and `pattribute(fallthrough)`.
Would it make sense to replace all `fallthrough` with `pattribute(fallthrough)` or is there a reason why using both variants?

bjosv:
Keeping pattributes in the portable library in case the usercode requires it.
The `fallthrough` define is no longer used internally and therefor not defined by default,
i.e PORTABLE_EXPOSE_ATTRIBUTES is not enabled by default.

No `implicit-fallthrough` build errors given in `./scripts/test.sh`

Fixes #247

### DIFF
--- a/src/compiler/semantics.c
+++ b/src/compiler/semantics.c
@@ -11,8 +11,6 @@
 #include <inttypes.h>
 #endif
 
-#include "flatcc/portable/pattributes.h" /* fallthrough */
-
 /* Same order as enum! */
 static const char *fb_known_attribute_names[] = {
     "",
@@ -525,7 +523,6 @@ static int analyze_struct(fb_parser_t *P, fb_compound_type_t *ct)
         member = (fb_member_t *)sym;
         switch (member->type.type) {
         case vt_fixed_array_type:
-            /* fall through */
         case vt_scalar_type:
             t = member->type.t;
             member->type.st = map_scalar_token_type(t);
@@ -538,7 +535,6 @@ static int analyze_struct(fb_parser_t *P, fb_compound_type_t *ct)
             member->size = size * member->type.len;
             break;
         case vt_fixed_array_compound_type_ref:
-            /* fall through */
         case vt_compound_type_ref:
             /* Enums might not be valid, but then it would be detected earlier. */
             if (member->type.ct->symbol.kind == fb_is_enum) {
@@ -702,8 +698,9 @@ static int process_struct(fb_parser_t *P, fb_compound_type_t *ct)
         switch (member->type.type) {
         case vt_fixed_array_type_ref:
             key_ok = 0;
-            fallthrough;
+            goto lbl_type_ref;
         case vt_type_ref:
+lbl_type_ref:
             type_sym = lookup_type_reference(P, ct->scope, member->type.ref);
             if (!type_sym) {
                 error_ref_sym(P, member->type.ref, "unknown type reference used with struct field", sym);
--- a/src/runtime/json_parser.c
+++ b/src/runtime/json_parser.c
@@ -141,11 +141,10 @@ descend:
         ++buf;
     }
     while (buf != end && *buf <= 0x20) {
-        /* Fall through comments needed to silence gcc 7 warnings. */
         switch (*buf) {
         case 0x0d: buf += (end - buf > 1 && buf[1] == 0x0a);
             /* Consume following LF or treating CR as LF. */
-            fallthrough;
+            ++ctx->line; ctx->line_start = ++buf; continue;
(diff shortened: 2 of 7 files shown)


bjosv:
Maybe you liked the label names like `lbl_fallthrough` better?

mikkelfj:
Well, on naming, in cases where fallthrough is at the very core, maybe fallthrough is the right label, but in general I tend to prefer names where it would also work if I moved the code around.

mikkelfj:
I'm still looking through the code, but one thing:
I think the label should be after the case label, not before. This both ensures that when moving code, it remains stable, and technically you still have fall through if you goto before the case label. And this is also how I have seen others suggest it.

src/runtime/json_parser.c
```diff
@@ -141,11 +141,11 @@ const char *flatcc_json_parser_space_ext(flatcc_json_parser_t *ctx, const char *
...
     while (buf != end && *buf <= 0x20) {
-        /* Fall through comments needed to silence gcc 7 warnings. */
         switch (*buf) {
         case 0x0d: buf += (end - buf > 1 && buf[1] == 0x0a);
             /* Consume following LF or treating CR as LF. */
-            fallthrough;
+            goto fallthrough;
+fallthrough:
         case 0x0a: ++ctx->line; ctx->line_start = ++buf; continue;
         case 0x09: ++buf; continue;
```
mikkelfj, line 150: Here case 0x09 need to be multiline and a label inserted between 0x0a: and ++ctx. Or perhaps better just copy the logic and avoid that mess.

mikkelfj:
Yeah, aside from the two comments I added, I think it looks fine. Let me know if you disagree.

bjosv:
I'll update, I've now seen some suggestion about having the label after the case.
I was thinking that if a new case was entered in the middle, or if a case was moved it wouldn't be a fallthrough anymore, more of a stepping over a case..

mikkelfj:
> I was thinking that if a new case was entered in the middle, or if a case was moved it wouldn't be a fallthrough anymore, more of a stepping over a case..

True, speaking generally and on this code in particular, often fall through is just a convenient placement and not truly a fall through, more like implementing a flow graph. Rather than trying to emulate fall through I think it is better to see at as a goto driven flow graph similar to goto error: labels. But it depends on the situation. In the case with falling numbers, fall through is at the very core of the logic.

src/runtime/json_parser.c
bjosv: Copied the logic.

src/compiler/semantics.c
```diff
@@ -702,7 +698,8 @@ static int process_struct(fb_parser_t *P, fb_compound_type_t *ct)
         switch (member->type.type) {
         case vt_fixed_array_type_ref:
             key_ok = 0;
-            fallthrough;
+            goto fallthrough;
+fallthrough:
         case vt_type_ref:
             type_sym = lookup_type_reference(P, ct->scope, member->type.ref);
```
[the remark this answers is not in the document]
bjosv: Done

mikkelfj:
OK if you are happy, I'm ready to merge.

bjosv:
It got a bit verboser, but I'm happy its according to recommendations.

PR #254, opened 2022-11-17
Enable to build using Clang 15

bjosv:
The diagnostics in Clang 15 is [stricter](https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics) and finds some issues.

- [-Werror,-Wstrict-prototypes]
  Using a strict prototype for some main()'s in the test code.

- [-Werror,-Wunused-but-set-variable]
  The variable 'present_id' is set but not used, so removed.

Some details: https://reviews.llvm.org/D122895


mikkelfj:
I have seen this main issue before. I generally just fix the main() call.
Are there other places that are less trivial to fix?

As to the present_id, I'd like to understood this better. That code wasn't just added for no reason.

bjosv:
Yes, I'm happy.

PR #260, opened 2023-01-15
Avoid false detection of Clang while using Clang-cl

taloz5:
When working with Clang-cl on MSVC, the compiler ID is set to Clang while it expects flag in MSVC style.
So need to check also `CMAKE_CXX_SIMULATE_ID` and not only `CMAKE_C_COMPILER_ID`


mikkelfj:
Thanks

PR #263, opened 2023-04-28
Install required gcc version in weekly CI for macOS

bjosv:
gcc-9 was removed from Githubs macOS runner via https://github.com/actions/runner-images/issues/7136
This PR adds a prepare step which installs the required gcc version.

[Here](https://github.com/Nordix/flatcc/actions/runs/4828087072) is a forced run.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -148,6 +148,9 @@ jobs:
         gcc-version: [9, 12]
     steps:
       - uses: actions/checkout@v3
+      - name: Prepare
+        run: |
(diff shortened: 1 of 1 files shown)


mikkelfj:
Excellent, thanks. It was bugging me a bit.

Generally, I think these build images are bit fragile. In the relatively short time we have had this GH build, two different images failed with missing compilers. One of them came back after a few weeks. GH support do say it is not their problem because it is community provided images ...

bjosv:
Yes, it's also annoying that the runner images have a version specified, but its not possible to pin that version in the workflows like how the actions can be pinned.

PR #264, opened 2023-05-30
add flatccrt_wdk

ggeenn:


mikkelfj:
Can you explain a bit behind about the motivation behind this?

While I am glad to see FlatCC being used in this context, I am not  convinced that the main repo should maintain a build for Wdk, notably where it isn't going to be tested regularly and where it is unlikely to benefit the majority of users, and which might require updates for future Windows versions.

Also, there are things like changing the main build using file globbing for runtime files which is not suitable as is since we want precise control over what we put in the library, regardless of what other files might be around.

Most users with special uses for runtime library simply copy the few runtime files into their own separate build. As to the main flatcc build, not runtime, I'm not sure it needs any Wdk specialization at all, but I haven't looked carefully.

So overall, I'd suggest maintaining you own separate build for projects like Wdk.

mikkelfj:
Closing this due to no follow up from author and due to the PR likely not being a good fit for the main project.

PR #266, opened 2023-10-04
Update the apt key for Intel compilers in weekly CI

bjosv:
The weekly CI job that builds with Intel compilers now downloads and uses the [recommended keyfile](https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-linux/2023-2/apt.html) for Intels APT repository.
This should fix failing weekly runs.

We also replace the deprecated `apt-key` command with `gpg  --dearmor` which is generally recommended.
`apt-key` has security flaws and will be removed after Ubuntu 22.04.
This change follows [Intels installation guide](https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-linux/2023-2/apt.html).

Logs from a forced run of this PR can be found [here](https://github.com/Nordix/flatcc/actions/runs/6407705818/job/17395072085).

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -112,8 +112,10 @@ jobs:
     steps:
       - name: Prepare
         run: |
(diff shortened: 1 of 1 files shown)


mikkelfj:
thanks

PR #273, opened 2024-02-02
Check for C standard version >= 23 for [[__fallthrough__]]

kraj:
This avoids warnings with clang-18+ e.g.

pprintint.h:256:6: error: [[]] attributes are a C23 extension [-Werror,-Wc23-extensions]

### DIFF
--- a/include/flatcc/portable/pattributes.h
+++ b/include/flatcc/portable/pattributes.h
@@ -57,7 +57,7 @@ extern "C" {
 
 
 /* https://en.cppreference.com/w/c/language/attributes/fallthrough */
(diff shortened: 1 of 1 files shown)


mikkelfj:
Thanks. I wonder if it is better to check for __has_c_attribute ?
https://clang.llvm.org/docs/LanguageExtensions.html#has-c-attribute

kraj:
> Thanks. I wonder if it is better to check for __has_c_attribute ? https://clang.llvm.org/docs/LanguageExtensions.html#has-c-attribute https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fc_005fattribute.html

it will still warn because the check will pass as clang extension even for pre c23 standard case

PR #275, opened 2024-03-14
CI: updates to Weekly build

bjosv:
- Fix warning "Node.js 16 actions are deprecated"
   Start using version `v2` of `actions-setup-cmake` which runs on Node.js 20.

- Add the latest `clang` and `gcc`,  the versions given by aminya/setup-cpp are:
    - gcc-13.1
    - clang-17.0.6

An [example](https://github.com/Nordix/flatcc/actions/runs/8268239669) run.

Note: an intermittent warning have been seen from the action `aminya/setup-cpp`, see https://github.com/aminya/setup-cpp/issues/234

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -182,7 +182,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Setup cmake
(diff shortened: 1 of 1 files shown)


mikkelfj:
thanks!

PR #276, opened 2024-04-09
Improve Github repository security

bjosv:
This PR updates our CI setup according to the secure software development best practices
recommended by the Open Source Security Foundation ([OpenSSF](https://openssf.org/)).
The overall goal is to strengthen the (supply chain) security posture.

The following changes are included:
- Pin the Github Action dependencies to hash.
When developing a CI workflow, it's common to version-pin dependencies (i.e. `actions/checkout@v4`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.
Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.
https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

- Set the default permission for CI workflows to only be able to read from the repository (scope: `contents`).
A compromised action will not be able to modify the repo or even steal secrets since all other permission-scopes are implicit set to "none", i.e. not permitted.
https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

- Add a [dependabot](https://docs.github.com/en/code-security/dependabot) which will perform weekly checks of the Github actions used in CI.
When a newer version is found a pull request is opened to suggest a lift.
https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool

For future reference, additional Github [guidelines](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) and info about [permissions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# Automated dependency updates.
+#
+# For configuration options see:
+# https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on: [push, pull_request]
 env:
   CTEST_OUTPUT_ON_FAILURE: 1
 
+permissions:
+  contents: read
+
 jobs:
   ubuntu-ninja-clang:
     name: Ubuntu (ninja, clang)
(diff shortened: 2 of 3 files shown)


mikkelfj:
Excellent and very timely wrt. xz ...

bjosv:
Yes, xz is somewhat an eye-opener..for me atleast.

These changes seems to improve OpenSSF scorecard from `Aggregate score: 4.2 / 10` to `Aggregate score: 6.5 / 10` (using public data), quite ok.

mikkelfj:
@bjosv Prophetic, look at the date:
https://www.youtube.com/watch?v=9qljpi5jiMQ&t=875s

bjosv:
Wow :) Thanks for the link, I didn't know the history with Azure Pipelines. So that's the reason for some weirdness..

PR #282, opened 2024-05-21
Replace `macos-11` runner in CI since it's deprecated

bjosv:
The macOS 11 runner image will soon be removed (2024-06-24) according to [github.blog](https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/#macos-11-deprecation-and-removal), so lets replace `macos-11` with `macos-12`. The gcc versions used in the CI job `macos-gcc` are also lifted since `gcc-9` wont run on `macos-12` (due to macosx-version-min).

This PR also adds the latest release of `clang` to CI.

A weekly run:
https://github.com/Nordix/flatcc/actions/runs/9172372095

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        clang-version: [5, 7, 9, 11, 13, 15, 17]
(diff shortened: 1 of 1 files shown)


mikkelfj:
Looks good thanks, and thanks for looking after dependabot!

PR #288, opened 2024-08-16
Use supported GCC versions in weekly CI for macOS

bjosv:
Since gcc@11 suddenly started to fail on runner `macos-14` due to a non-matching default Xcode version;
lets only use supported versions of GCC in the macOS weekly runs.
These GCC versions are more likely verified to work fine on runners.

Only the 3 latest versions of GCC are supported, i.e. GCC 12, 13 and 14 according to
https://github.com/actions/runner-images/issues/10213 and https://github.com/actions/runner-images#software-and-image-support .

Manual run: https://github.com/Nordix/flatcc/actions/runs/10418446489

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -160,7 +160,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gcc-version: [11, 14]
+        gcc-version: [12, 14] # Github supports the 3 latest major versions.
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
       - name: Prepare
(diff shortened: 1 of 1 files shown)


mikkelfj:
Thanks a lot!

mikkelfj:
I also run on main and it passes. But there is a clang 15 annotated warning:
https://github.com/dvidelabs/flatcc/actions/runs/10419893918


Clang 15
Falling back to LLVM version 18.1.8 -ubuntu-18.04 for the Ubuntu.

bjosv:
Yes, very weird, CMake says its using the expected version at least
```
2024-08-16T12:29:08.8225520Z -- The C compiler identification is Clang 15.0.6
2024-08-16T12:29:08.9566399Z -- The CXX compiler identification is Clang 15.0.6
```
I'll try to see where in the action the warning comes from

Issue #287, opened 2024-08-14
Alignment (probably) shouldn't be required for empty vectors


bkietz:
Flatcc's `verify_vector` requires that vectors be properly aligned even when the vector is empty. The consumers in https://github.com/google/flatbuffers do not enforce this and moreover the producers will sometimes emit an empty vector which they didn't bother to align.

(Encountered in the context of arrow IPC, which [uses flatbuffers to store metadata](https://arrow.apache.org/docs/format/Columnar.html#encapsulated-message-format)) A minimal example: https://gist.github.com/bkietz/6150b9aae6d1bea0c85c0527d7fb04d6

mikkelfj:
@bkietz thanks for making a note of this. I am not able to dig deeply into this at the moment, I am thrilled that flatcc is being used with Arrow, and I did how some feedback to one of the Arrow designers back then. I have worked a bit with the Arrow format in Julia and Go.

That said, I disagree about alignment. You could get away with it, sort of, for empty vectors, but the (empty) data still needs to be aligned to at least 32-bits because of the length header. Furthermore pointer types in some languages, notably C, are required to be aligned, at the very least if dereferenced, but even if not, who knows what aggressive optimizations a compiler might do. In C vectors can be accessed via raw pointers iff the user takes responsibility for endian conversion, or the platform is known little endian. Therefore there are functions that are expected to return aligned pointers even if the length is empty.

Non-alignment can also lead to unexpected errros such as a user doing pointer math to get the header field, even if the user is not supposed to that directly, since the binary format is well documented. Even a compiler might do that, though unlikely.

As for the google/flatbuffers consumers: No consumers I am aware of enforce alignment, notably, it is perfectly valid to read from an unaligned buffer, e.g. from a network buffer, iff you are sure you platform arch can handle it, such as x86 and some, but not all ARM. Verification is different though. Googles C++ verifier did not verify alignment very aggressively early AFAIR, but that does not mean it should not do that, only that it did not implement it.

From a practical perspective, it would also be an odd exception in the verifier to first check and a fail, then add branch logic to check if it in fact happened to  be an empty vector, and doing it before would be a an unwanted performance penality, and still a complication.

Therefore I suggest you try to address the issue at the source.

mikkelfj:
On a related note:

I would like the verifier to be able to accept unaligned buffers but still verify that they are properly aligned relative to buffer start. This would be useful for verify buffers in unaligned network buffers without having to make a copy.

[...]

PR #289, opened 2024-08-16
Add FLATCC_ENFORCE_ALIGNED_EMPTY_VECTORS

bkietz:
An option is added which, if enabled, causes verifier to only require empty vectors be aligned to 4 bytes.

discussion at https://github.com/dvidelabs/flatcc/issues/287

### DIFF
--- a/include/flatcc/flatcc_flatbuffers.h
+++ b/include/flatcc/flatcc_flatbuffers.h
@@ -48,6 +48,23 @@ extern "C" {
 #define FLATBUFFERS_WRAP_NAMESPACE(ns, x) ns ## _ ## x
 #endif
 
+#include <stddef.h>
+
+/*
+ * Produce a pointer with maximum alignment to use as
+ * a placeholder when FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS
+ * is specified.
+ */
+static inline const void* max_aligned_ptr(void)
+{
+#ifdef __cplusplus
+    static const max_align_t m{};
+#else
+    static const max_align_t m;
+#endif
+    return &m;
+}
+
 #endif /* flatcc_flatbuffers_defined */
 
 #ifdef __cplusplus
--- a/include/flatcc/flatcc_flatbuffers.h
+++ b/include/flatcc/flatcc_flatbuffers.h
@@ -48,23 +48,6 @@ extern "C" {
 #define FLATBUFFERS_WRAP_NAMESPACE(ns, x) ns ## _ ## x
 #endif
 
-#include <stddef.h>
-
-/*
- * Produce a pointer with maximum alignment to use as
- * a placeholder when FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS
- * is specified.
- */
-static inline const void* max_aligned_ptr(void)
-{
-#ifdef __cplusplus
-    static const max_align_t m{};
-#else
-    static const max_align_t m;
-#endif
-    return &m;
-}
-
 #endif /* flatcc_flatbuffers_defined */
 
 #ifdef __cplusplus
--- a/include/flatcc/flatcc_rtconfig.h
+++ b/include/flatcc/flatcc_rtconfig.h
@@ -66,12 +66,7 @@ extern "C" {
 /*
  * Some producers allow empty vectors to be misaligned.
  * The following setting will cause the verifier to check for an
- * empty vector before checking alignment. This option will also
- * ensure accesses to a vector field never materializes misaligned
- * pointers.
- *
- * NOTE: enabling this means empty vectors will not have distinct
- * pointer identity.
+ * empty vector before checking alignment of the vector's elements.
  */
 #if !defined(FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS)
 #define FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS 0
--- a/include/flatcc/flatcc_rtconfig.h
+++ b/include/flatcc/flatcc_rtconfig.h
@@ -65,11 +65,12 @@ extern "C" {
 
 /*
  * Some producers allow empty vectors to be misaligned.
- * The following setting will cause the verifier to check for an
- * empty vector before checking alignment of the vector's elements.
+ * The following setting will cause the verifier to require the index 0
+ * position to be element aligned even if the vector is empty (otherwise that
+ * position is only required to be aligned to the preceding size field).
  */
-#if !defined(FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS)
-#define FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS 0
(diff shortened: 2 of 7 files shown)


mikkelfj (commented):
That was fast. I don't entirely agree with updating the reader as per my comments inline, but I'm impressed you were able to dig that deep into the code so fast.

I think we should focus on getting the verifier online. We could also do a code review to understand the implications of unligned vectors for the actual code. 

Maybe we should rename the flat FLATCC_VERIFIER_TOLERATE_... to indicate that this is then verfier specific.

test/load_test/load_test.c
```diff
@@ -136,6 +136,8 @@ int main(int argc, char *argv[])
         create_root_monster(B);
         flatcc_builder_copy_buffer(B, buffer, size);
         mon = ns(Monster_as_root(buffer));
+        char const *name = ns(Monster_name(mon));
```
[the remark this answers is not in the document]
bkietz: I was trying to debug what I've broken in string field accessing. It would not be included in this PR

src/compiler/codegen_c_reader.c
```diff
@@ -618,10 +620,34 @@ static void gen_helpers(fb_output_t *out)
         "    return 0;\\\n"
         "}\n",
         nsc, nsc, nsc, nsc, nsc);
+#if FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS
     fprintf(out->fp,
-        "#define __%svector_field(T, ID, t, r) __%soffset_field(T, ID, t, r, sizeof(%suoffset_t))\n"
-        "#define __%stable_field(T, ID, t, r) __%soffset_field(T, ID, t, r, 0)\n",
+        "#include <stdio.h>\n"
```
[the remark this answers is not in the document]
bkietz: This is more debug code. I'll remove it

include/flatcc/flatcc_flatbuffers.h
```diff
@@ -48,6 +48,23 @@ extern "C" {
...
 #endif
 
+#include <stddef.h>
+
+/*
+ * Produce a pointer with maximum alignment to use as
+ * a placeholder when FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS
+ * is specified.
+ */
+static inline const void* max_aligned_ptr(void)
```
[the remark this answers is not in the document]
bkietz: Since it seems that avoiding *materialization* of misaligned pointers (and in general modifying the reader) is not essential for this PR, I'll just revert all those changes.

src/runtime/verifier.c
```diff
@@ -276,12 +262,18 @@ static inline int verify_string(const void *buf, uoffset_t end, uoffset_t base,
...
 static inline int verify_vector(const void *buf, uoffset_t end, uoffset_t base, uoffset_t offset, uoffset_t elem_size, uint16_t align, uoffset_t max_count)
 {
-    uoffset_t n;
-
-    verify(check_aligned_header(end, base, offset, align), flatcc_verify_error_vector_header_out_of_range_or_unaligned);
+    verify(check_header(end, base, offset), flatcc_verify_error_vector_header_out_of_range_or_unaligned);
     base += offset;
-    n = read_uoffset(buf, base);
+
+    uoffset_t n = read_uoffset(buf, base);
```
mikkelfj, line 268: Flatcc declares variables before statements, so please keep uoffset_t n;

include/flatcc/flatcc_rtconfig.h
```diff
@@ -63,6 +63,14 @@ extern "C" {
 #define FLATCC_TRACE_VERIFY 0
 #endif
 
+/*
+ * Some producers allow empty vectors to be misaligned.
+ * The following setting will cause the verifier to check for an
+ * empty vector before checking alignment of the vector's elements.
+ */
```
mikkelfj, line 70: The following setting will not require the index 0 position to be element aligned iff the vector is empty, but it must still be aligned to the preceding size field.

(note: I am not 100% if C++ even aligns the size field correctly in this case, but I don't think I have seen evidence to the contrary. If the size field is not aligned, the problem is worse, both in verifier, and the reader also needs to be handled with more care).

include/flatcc/flatcc_rtconfig.h
```diff
@@ -63,6 +63,14 @@ extern "C" {
...
+#if !defined(FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS)
+#define FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS 0
+#endif
```
mikkelfj, line 74: I think we may have to set this to 1 and flip the CMake setting.

src/runtime/verifier.c
```diff
@@ -276,12 +262,18 @@ static inline int verify_string(const void *buf, uoffset_t end, uoffset_t base,
...
     base += offset_size;
+
```
mikkelfj, line 270: Add comment like:
/* This is due to incorrect buffers from other builders than cannot easily be ignored. */

src/runtime/verifier.c
```diff
@@ -276,12 +262,18 @@ static inline int verify_string(const void *buf, uoffset_t end, uoffset_t base,
...
+#if FLATCC_TOLERATE_MISALIGNED_EMPTY_VECTORS
+    align = n == 0 ? uoffset_size : align;
+#endif
+    align = align < uoffset_size ? uoffset_size : align;
```
mikkelfj, line 274: You could place this in an #else branch. Then we have zero extra performance overhead.

mikkelfj (commented):
I think this looks good, a part from a few minor comments:

src/runtime/verifier.c
bkietz: If `align = 1, n = 2` (for a vector of two bytes) then `align = n == 0 ? offset_size : align` will yield alignment 1. I think we need to keep the independent clamp to offset_size

include/flatcc/flatcc_rtconfig.h
bkietz: In that case, I should probably rename the option to `FLATCC_REFUSE_MISALIGNED_EMPTY_VECTORS` and leave *its* default at 0

src/runtime/verifier.c
mikkelfj: It doesn't matter because any alignment < offset_size will be overridden by the header check.

include/flatcc/flatcc_rtconfig.h
mikkelfj: FLATCC_ENFORCE_ALIGNED_EMPTY_VECTORS

src/runtime/verifier.c
mikkelfj: hmm but would also be true in the other  case, so why is the line there?

We could remove it and it should be ok, but to be safe, we could also have:

verify(!(base & ((align - 1u) | (uoffset_size - 1u))

mikkelfj:
Yeah, I think that is good, aside from the PR title. When you are ready I'll squash and merge.

PR #296, opened 2024-09-18
Revert "Bump aminya/setup-cpp from 0.39.0 to 0.41.0"

bjosv:
Investigated the issue with the weekly CI runs failing and created this issue:  https://github.com/aminya/setup-cpp/issues/295
Lets revert this version to the older version that used a more accepting client while downloading llvm packages.
Maybe the issue is temporary but lets use the older version until I can make sure, or the issue is fixed in setup-cpp.

Reverts dvidelabs/flatcc#294


mikkelfj:
Thanks.

PR #308, opened 2024-10-25
consider length of string keys

mdrossmiller:
take length of keys in the str_set hash into account so keys which are perfect sub-strings do not match incorrectly

### DIFF
--- a/external/hash/str_set.c
+++ b/external/hash/str_set.c
@@ -47,7 +47,7 @@ DEFINE_HASH_TABLE(str_set)
  */
 static inline int ht_match(const void *key, size_t len, str_set_item_t item)
 {
-    return len == strlen(item) && strncmp(key, item, len) == 0;
+    return strncmp(key, item, len) == 0 && item[len] == '\0';
 }
 
 static inline const void *ht_key(str_set_item_t item)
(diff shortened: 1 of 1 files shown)


external/hash/str_set.c
```diff
@@ -47,7 +47,7 @@ DEFINE_HASH_TABLE(str_set)
  */
 static inline int ht_match(const void *key, size_t len, str_set_item_t item)
 {
-    return strncmp(key, item, len) == 0;
+    return len == strlen(item) && strncmp(key, item, len) == 0;
 }
```
mikkelfj, line 51: return strncmp(key, item, len) == 0 && item[len] == '\0';
?

mdrossmiller: I applied your suggestion

mikkelfj:
thanks

mikkelfj:
Note: this is a low impact issue unless str_set is used separately from flatcc as is. See also CHANGELOG note. Essentially str_set is used to detect if a header file has already been included in generated output. A false positive of a prefix match is unlikely due to hashing, and if it happens, worst case the end code might not compile without manual inclusion of a file assumed to be already included.

PR #312, opened 2024-11-20
runtime/CMakeLists.txt: Use target_include_directories

midokura-xavi92:
Otherwise, include directories are not propagated to targets calling `target_link_libraries()`.

### DIFF
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -1,7 +1,3 @@
-include_directories (
-    "${PROJECT_SOURCE_DIR}/include"
-)
-
 add_library(flatccrt
     builder.c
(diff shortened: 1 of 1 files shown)


mikkelfj:
I bumped CMake to 3.16 and fixed a lot of old warnings on custom builds, which makes this PR hopefully no longer needed, but if you have suggestions on top of that, we are in a better position to deal with it now.

Bumped CMake to 3.16 in this commit: [385c27b](https://github.com/dvidelabs/flatcc/commit/385c27b23236dff7ad4fa35c59fd4f9143dcaae6)

Version 3.16 still allows Appveyor to test MSVC 2015 and should be recent enough.

PR #314, opened 2024-11-25
bump the minimum version of cmake

inthehack:
This commit bumps the minimum version of CMake in CMakeLists.txt files.

The version of CMake in Ubuntu LTS is 3.28 and in CentOS is 3.26. Then, the version of CMake is set to 3.25 arbitrary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,5 @@
-# Ubuntu 14.04 (Trusty)
-cmake_minimum_required (VERSION 2.8.12.2)
-# Centos 7
-#cmake_minimum_required (VERSION 2.8.11)
-#cmake_minimum_required (VERSION 2.8)
+# CMake minimal version.
+cmake_minimum_required (VERSION 3.25)
 
 # Experimental for generating compile_commands.json so editors with
 # clangd language server support can use it. Symlink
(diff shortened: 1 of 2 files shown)


mikkelfj:
Sorry for the last answer. There is a slow long term plan to bump CMake, but it was indeed tied up to Ubuntu LTS and the desire to support older Ubuntu.

@bjornharrtell We are just right now bumping to a more recent LTS in CI build, but other users might not.

Therefore I do not currently have a clear view on what is the safe CMake version, but there are some issues where this has been discussed in the past. But I do recognize that the current CMake version is outdated.

mikkelfj:
Also note that AppVeyor currently fails this PR due it running: cmake version 3.16.2
We also need to figure out a plan for migrating Windows builds to GH actions, but there is a lack of support for older MSVC compilers here.

mikkelfj:
BTW: I am now getting some strange warnings in CMake on MacOS related to add_custom_command that likely needs to be dealt with.

CMake Warning (dev) at test/monster_test_cpp/CMakeLists.txt:13 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): DEPENDS.
  
  CMake Warning (dev) at test/monster_test/CMakeLists.txt:10 (add_custom_command):
  Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given.  Assuming
  POST_BUILD to preserve backward compatibility.

PR #315, opened 2024-11-25
fix argument type in fprintf

inthehack:
This commit fixes the argument type of member->type.len, which is a uint32_t and not an int.


mikkelfj:
Thanks, this is a little belated, but there you go. If the type is indeed iuint32_t as you claim, the field should indeed not be %d, but %ld is not right either. It needs to be `%"PRIu32"`.

If you have moved on, I will look into this eventually.

mikkelfj:
Fixed in d8ddb2c

PR #321, opened 2025-03-10
flatbuffer reading support for windows kernelland

Chuccle:
Tinkered a bit to allow reading of flatbuffers without stdlib, obviously you'll need to define the stdint types and `#defines` in your build environment.

Very much a new contributor to open source so apologies in advance

### DIFF
--- a/include/flatcc/flatcc_endian.h
+++ b/include/flatcc/flatcc_endian.h
@@ -45,7 +45,7 @@ extern "C" {
  * configuration.
  */
 
-#ifndef UINT8_t
+#ifndef UINT8_MAX
 #include <stdint.h>
 #endif
 
--- a/include/flatcc/portable/pstdint.h
+++ b/include/flatcc/portable/pstdint.h
@@ -197,7 +197,7 @@
  *  do nothing else.  On the Mac OS X version of gcc this is _STDINT_H_.
  */
 
-#if ((defined(_MSC_VER) && _MSC_VER >= 1600) || (defined(__STDC__) && __STDC__ && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || (defined (__WATCOMC__) && (defined (_STDINT_H_INCLUDED) || __WATCOMC__ >= 1250)) || (defined(__GNUC__) && (__GNUC__ > 3 || defined(_STDINT_H) || defined(_STDINT_H_) || defined (__UINT_FAST64_TYPE__)) )) && !defined (_PSTDINT_H_INCLUDED)
+#if ((defined(_MSC_VER) && _MSC_VER >= 1600) || (defined(__STDC__) && __STDC__ && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || (defined (__WATCOMC__) && (defined (_STDINT_H_INCLUDED) || __WATCOMC__ >= 1250)) || (defined(__GNUC__) && (__GNUC__ > 3 || defined(_STDINT_H) || defined(_STDINT_H_) || defined (__UINT_FAST64_TYPE__)) )) && !defined (_PSTDINT_H_INCLUDED) && !defined (NTDDI_VERSION)
 #include <stdint.h>
 #define _PSTDINT_H_INCLUDED
 # if defined(__GNUC__) && (defined(__x86_64__) || defined(__ppc64__)) && !(defined(__APPLE__) && defined(__MACH__))


mikkelfj:
Thanks.
Can you explain the rationale and the actual use case?


A few notes:

1. the portably sub library cannot have any flatcc specific references. the include/flatcc/flatcc.h (AFAIR) acts as a middle layer when flagging in FLATCC_ features and translates them to portable features and includes files.

2. I am slowly (mostly new projects) moving towards not using the UINT8_MAX check since it is only a very old MSVC that really needs it. But this goes a bit against your approach.

3. I already prepared for some of you are doing: https://github.com/dvidelabs/flatcc/tree/master/include/flatcc/portable/include

4. I'm not sure one can just no get rid of stdlib.h. At least that needs to be tested. There are things like allocation and size_t, and what not, I don't recall exactly where what goes. 

5. As per the above, it is "easy" to get rid of stdint.h. Getting rid of stdlib.h I'm not so sure.

6. Users have already compiled with newlib and similar limited libraries. That is one reason all the allocation functions are abstracted in include/flatcc. You can try search for newlib in issues or PRs.

mikkelfj:
Oh, and BTW: since certain transformer related events, these bullet lists are getting out of hand.

Chuccle:
> Thanks. Can you explain the rationale and the actual use case?
> 
> A few notes:
> 
> 1. the portably sub library cannot have any flatcc specific references. the include/flatcc/flatcc.h (AFAIR) acts as a middle layer when flagging in FLATCC_ features and translates them to portable features and includes files.
> 2. I am slowly (mostly new projects) moving towards not using the UINT8_MAX check since it is only a very old MSVC that really needs it. But this goes a bit against your approach.
> 3. I already prepared for some of you are doing: https://github.com/dvidelabs/flatcc/tree/master/include/flatcc/portable/include
> 4. I'm not sure one can just no get rid of stdlib.h. At least that needs to be tested. There are things like allocation and size_t, and what not, I don't recall exactly where what goes.
> 5. As per the above, it is "easy" to get rid of stdint.h. Getting rid of stdlib.h I'm not so sure.
> 6. Users have already compiled with newlib and similar limited libraries. That is one reason all the allocation functions are abstracted in include/flatcc. You can try search for newlib in issues or PRs.

Thanks for getting back to me. A lot of useful stuff to take away.

I've got a windows driver and wanting to deserialize a flatbuffer received over a http kernel socket connection.

The windows kernel runtime conflicts and isn't compatible with it's userland counterpart and doesn't contain a good portion of it.

This PR was me following the '#include' list and guarding around the problematic headers.

I didn't spot the portable header, that might be the key to solving this elegantly, will see if I can sort the issue that way.

Chuccle:
Verifier.c when compiled in my environment seems to be working as a straight drop in which is brill 😃

mikkelfj:
Wrt. Kernel driver, I should warn you that flatbuffers are not inherently safe. Malformed buffers can be overrun and even a valid verifier can be denial of service attack'ed, so you only want to do this if you have full control over the data. The safety of the verifier can be improved by adding additional checks for long runtimes, but it could still be slow to handle valid cases.

There was a similar discussion with Linux kernel where the idea was abandoned.

On the pstdalign header, you also just add `/DPORTABLE_HAS_INCLUDE_STDALIGN=0` in you build system since NTDDI is a rather specific case.

I don't quite like how you made the change since that section is supposed to do just one thing and it is hard enough to reason about as is. However, I also don't have a better suggestion other than wondering if MSC_VER might be a better test elsewhere?

stdint.h is probably fine, I can't see it doing much harm except that I vary occasionally import from the original source which is where I have earlier contributed changes.

Chuccle:
> Wrt. Kernel driver, I should warn you that flatbuffers are not inherently safe. Malformed buffers can be overrun and even a valid verifier can be denial of service attack'ed, so you only want to do this if you have full control over the data. The safety of the verifier can be improved by adding additional checks for long runtimes, but it could still be slow to handle valid cases.
> 
> There was a similar discussion with Linux kernel where the idea was abandoned.
> 
> On the pstdalign header, you also just add `/DPORTABLE_HAS_INCLUDE_STDALIGN=0` in you build system since NTDDI is a rather specific case.
> 
> I don't quite like how you made the change since that section is supposed to do just one thing and it is hard enough to reason about as is. However, I also don't have a better suggestion other than wondering if MSC_VER might be a better test elsewhere?
> 
> stdint.h is probably fine, I can't see it doing much harm except that I vary occasionally import from the original source which is where I have earlier contributed changes.

I know this probably isn't the most appropriate medium regarding security questions but regarding malformed buffers, presumably verifying a buffer as root involves some level of offset checking to prevent access violations, that's the most important thing for me.

As for DoS, I want to say that I will refuse stupidly large payloads based off content-length and my own internal buffer allocations. Might be a bit naïve for now but it **should** be enough for now

mikkelfj:
https://github.com/dvidelabs/flatcc/blob/master/doc/security.md

Quoting from securirty doc:

> A buffer can point to the same data from multiple places. This is known as a DAG. The verifier rejects cycles that could lead to infinite loops during application traversal but does permit DAGs. For normal use DAGs are safe but it is possible to maliciously construct a buffer with a long vector where all elements points to a table that also has a vector of a similar nature. After a few levels, this can lead to a finite but exponentially large number of places to visit. The current FlatCC verifier does not protect against this but Googles flatc compiler has a verifier the limits the number of visited tables.

mikkelfj:
If you do this in a thread with a timeout you should be safe.

Chuccle:
Thanks for all the info. I realise it's a bit of a RTFM moment on my part considering you've already gone to the trouble of making a security guide lol

PR #322, opened 2025-03-11
CI updates

bjosv:
- Replace ubuntu 20.04 since the runner will be removed 2025-04-01.
- Use the last release of `icc` to be able to update the ubuntu runner.
- Use available gcc/clang versions in weekly CI runs (gcc-7 removed).
- Bump aminya/setup-cpp from 0.40.0 to 1.1.1

Weekly run looks fine:
https://github.com/Nordix/flatcc/actions/runs/13786429513


mikkelfj:
thanks!

PR #323, opened 2025-03-14
Sync flatcc_verify_typed_buffer_header_with_size parameter name vs. definition

MichaelChirico:
https://github.com/dvidelabs/flatcc/blob/b50e43dd1269b91203be8b5b185923d841c93e70/src/runtime/verifier.c#L511

As flagged by `clang-tidy`:

https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html

---

The alternative is to change the name in the definition, which requires a bigger diff. The function is short enough that there's not much difference either way, though.


mikkelfj:
thanks

PR #338, opened 2025-05-22
fix(grisu3): GCC15 Complaining about non-null terminated string

fedefrancescon:
GCC15 started issuing a warning for non-null terminated string initiaziotions (`unterminated-string-initialization`).  
Avoiding the error by letting the compiler append the NULL

### DIFF
--- a/external/grisu3/grisu3_print.h
+++ b/external/grisu3/grisu3_print.h
@@ -183,7 +183,7 @@ static int grisu3_i_to_str(int val, char *str)
 
 static int grisu3_print_nan(uint64_t v, char *dst)
 {
-    static char hexdigits[16] = "0123456789ABCDEF";
+    static char hexdigits[] = "0123456789ABCDEF";
     int i = 0;
 
     dst[0] = 'N';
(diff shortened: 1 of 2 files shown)


fedefrancescon:
There are similar issues around:
- https://github.com/Mbed-TLS/mbedtls/issues/9944
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117178
- https://github.com/micropython/micropython/issues/16692

While this costs 1byte, might not be an issue at all.  
For sure there are other possibile solutions if needed

mikkelfj:
Hi, sorry for not getting back earlier.

Please note other PR as well

https://github.com/dvidelabs/flatcc/pull/347

This should be an easy fix, but we need to make it portable. I'm not sure setting length is not enough as per my comment in other PR.

fedefrancescon:
Yes, this works fine with GCC-15!

Apart from this specific issue, would be interesting to add appveyor tests?

mikkelfj:
Ah, you mean GCC 15 tests?

Appveyor only runs older Windows builds, github actions currently runs up to GCC 14 weekly.
@bjosv will probably update when the time is right.

PR #343, opened 2025-06-12
Fail tests when UBSan finds issues

bjosv:
We have found that there are a couple of issues when building with the UndefinedBehaviorSanitizer.
Some checkers in the UBSan arsenal are recoverable and just gives a printout to stderr, i.e. ctest does not fail.
Adding `-fno-sanitize-recover=all` triggers tests to exit instead, and issues are visible.

While enabling this a couple issues surfaces, like unaligned access warnings in 10 test binaries, and most are fixed by a change in `pprintint.h` and `punaligned.h`.

There is an open question about how to solve the issue in `monster_test.c`, I'm not sure if there is proper fix.
Any feedback/ideas?

Weekly test run [OK](https://github.com/Nordix/flatcc/actions/runs/15612083989).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,12 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang" AND NOT "${CMAKE_CXX_SIMULATE_ID}" STREQ
         if (CMAKE_BUILD_TYPE MATCHES Debug)
             set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined")
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined")
+            # Don't recover from found issues, we want tests to exit. We then need `__has_feature`
+            # which was added in clang-8, see https://github.com/llvm/llvm-project/commit/b32d404
+            if (NOT CMAKE_C_COMPILER_VERSION VERSION_LESS 8)
+                set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-sanitize-recover=all")
+                set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-sanitize-recover=all")
+            endif()
         endif()
     endif()
     # Suppress warning relaxed in clang-6, see https://reviews.llvm.org/D28148
--- a/include/flatcc/portable/pprintint.h
+++ b/include/flatcc/portable/pprintint.h
@@ -102,7 +102,9 @@ static int print_int(int n, char *p);
 static int print_long(long n, char *p);
 
 
-#if defined(__i386__) || defined(__x86_64__) || defined(_M_IX86) || defined(_M_X64)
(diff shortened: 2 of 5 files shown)


mikkelfj:
Thanks for digging into this.

I'll need to look a bit closer to better understand the issue.

In general we don't want #if whatever, if can be avoided, except for dedicated portable/* files.

punligned.h is the ultimate reference so clearly this need to work. I'm not sure using sanitizer flags is right here, rather we need to understand the core issue and maybe adapt to the problematic platform using another branch in the file perhaps.

pprintint.h really should be using punaligned, but that might be an unnecessary dependency, but it depends on what we can come up with.

The general solution is to use memcpy as you mention, but it is not always optimized. Once did made a larger effort to ensure all runtime used memcpy, but it was involved, and resulted in some poor performance occasionally, but that was a while back.

I don't have any answers yet, and need to look into it a bit more. I'm also not very familiar with sanitizer thingy.

mikkelfj:
Wrt. monstertest:

I think the problem may be that we reuse the pointer after changing the underlying content:

    vec3 = ns(Vec3_as_root(buffer));
    /* Convert buffer to native in place - a nop on native platform. */
    v = (ns(Vec3_t) *)vec3;
    ns(Vec3_from_pe(v));
    
    
I don't recall the exact interface of Vec3_from_pe(v) - whether it needs a typed pointer or not, but we need a new pointer cast from char * or maybe void * before we attempt to access the data.

  // declare v_pe earlier ...
  v_pe = (ns(Vec3_t) *)vec3;
  ns(Vec3_from_pe(v_pe));
  v  = (nv(Vec3_t) *)(char *)vec3;
  
  I don't have test setup handy atm.

mikkelfj:
wrt printint:

I think we should just use memcpy here, it's an isolated case and probably fast anywhere JSON printer performance matters.

mikkelfj:
som thoughts:

- maybe we should add mem_copy_word_2 explicitly since we likely won't get around to token pasting optimizations. That way we can better control and fix potential loss of performance.

- even if we don't use punaligned much, I consider the portable library a goto reference, if not used directly then to copy snippets from, clearly flatcc use case is primary.

- I don't think we should assume anything about stack alignment at all. That sounds like a bug, and you seem to have fixed it.
New users often get it wrong with char buf[BUFSIZE] to hold a flatbuffer, which is of course not aligned, but it seems like I also made that mistake. Note that instead of alignas(16) I think it is also possible to alignas(<type>) which might be more appropriate for struct buffers, but alignas(16) is also fine as long as the struct doesn't align more than that. The type version should be more robust if the struct were ever to be modified.

Background:
C11 stdalign (essentially alignas) is used widely in generated code, and since at the time, it was not widely supported, pstdalign provides the abstraction necessary. This should be used in any stack allocatoin that needs to be aligned:
https://github.com/dvidelabs/flatcc/blob/be/include/flatcc/portable/pstdalign.h

- on json: you mention temporary change wrt.:
 if (n >= 8 && ((uintptr_t)buf % 8) == 0) {
 
 sure, but for production the %8 test would defeat any kind of optimization attempted.

Issue #345, opened 2025-06-17
Verifier on buffer with size prefix and file identifier reads file identifier at the wrong offset


ovenystas-mp:
Hi,

I'm just starting to use FlatBuffers and I noticed this bug.

In _verifier.c_, functions _flatcc_verify_buffer_header_with_size()_ and _flatcc_verify_typed_buffer_header_with_size()_ doesn't add the size_prefix offset when reading the file identifier. This results in that it reads the file identifier at the wrong place and verifier falsely fails with error flatcc_verify_error_identifier_mismatch.

I suggest this code change in both functions:

From:
`id = read_thash(buf, offset_size);`

To:
`id = read_thash(buf, 2 * offset_size);`

PR #346, opened 2025-06-17
Fix missing offset in verifier

ovenystas-mp:
Fix for issue #345

An issue when buffer size prefix and file identifier are used.


mikkelfj:
Thanks, finally got around to validate the PR.

The affected functions are:

```
flatcc_verify_buffer_header_with_size
flatcc_verify_typed_buffer_header_with_size
```

PR #349, opened 2025-07-09
cmake: Set minimum required version to 3.5 for CMake 4+ compatibility

alperak:
Fix:

| CMake Error at CMakeLists.txt:2 (cmake_minimum_required):
|   Compatibility with CMake < 3.5 has been removed from CMake.
|
|   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
|   to tell CMake that the project requires at least <min> but has been updated
|   to work with policies introduced by <max> or earlier.
|
|   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
|
|
| -- Configuring incomplete, errors occurred!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Ubuntu 14.04 (Trusty)
-cmake_minimum_required (VERSION 2.8.12.2)
+cmake_minimum_required (VERSION 3.5)
 # Centos 7
 #cmake_minimum_required (VERSION 2.8.11)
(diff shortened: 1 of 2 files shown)


mikkelfj:
Thanks, I'm hesitating a bit on the decision. The current version is too old, but new version might not work on some platforms, and I think there are things in the current CMake files that need to be fixed before bumping.

There are also some other PR/issues that I need to consider.

Any sugg. are welcome.

PR #351, opened 2025-07-16
CI updates

bjosv:
- Remove [deprecated](https://github.com/actions/runner-images/issues/12045) Github runner `windows-2019` and add `windows-2025` in weekly CI.
- Don't install the already installed `ninja` on `macos` runners.
- Add `clang-20` to weekly CI and..
- Disable new warning when building `monster_test_cpp`.
```
      monster.c:294:5: warning: 'main' should not be 'extern "C"' [-Wmain]
      294 | int main(int argc, char *argv[])
          |     ^
``` 

[Weekly](https://github.com/Nordix/flatcc/actions/runs/16322137465) runs pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,9 +42,6 @@ jobs:
     name: macOS
     runs-on: macos-14
     steps:
-      - name: Prepare
-        run: |
-          brew install ninja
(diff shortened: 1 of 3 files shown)


mikkelfj:
Thanks. Yes, extern main is probably best fixed like that for now. We cannot really break up the original sample just because of that.

PR #352, opened 2025-07-23
Add gcc-15 and big-endian runs in weekly CI

bjosv:
- Add gcc-15 to weekly CI.
  Currently gcc-15 is only provided in ubuntu's `25.04`. 
  Mentioned in #338.

- Add a job to weekly CI that builds and runs on a big-endian platform (s390x via QEMU).

A run on [weekly](https://github.com/Nordix/flatcc/actions/runs/16467686523).

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -91,6 +91,24 @@ jobs:
           scripts/initbuild.sh make-concurrent
           scripts/test.sh
 
+  gcc-15:
+    name: GCC 15
+    runs-on: ubuntu-24.04
+    container:
+      image: ubuntu:25.04 # Provides gcc-15
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Prepare
+        run: |
+          apt update && apt-get install -y gcc-15 g++-15 cmake
+      - name: Build and run tests
+        env:
+          CC: gcc-15
+          CXX: g++-15
+        run: |
+          scripts/initbuild.sh make-concurrent
+          scripts/test.sh
+
   gcc-32bit:
     name: GCC 32bit
     runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


mikkelfj:
Thanks, great to see big endian covered.

FYI: probably not important but there was this warning on the QEMU build in weekly.

- FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/s390x" (line 1)

12min is long build time, but I guess that is what it takes.

FYI:
A long time ago we did a native big endian build for esoteric hardware. It is not binary compatible with flatbuffers which is little endian on all platforms, but rather used a native big endian platform useful for interprocess com. on such HW.
buffer fileidentier like MONS (for monster test became SNOM (the reverse) just to identify the difference):

https://github.com/dvidelabs/flatcc/tree/be

there is also a befloat branch, but I don't recall the point of it.

mikkelfj:
Other issue:

I think we need to add a Debian stable build, maybe in conjunction with the CMake min version build.

I'm looking into which CMake version to upgrade to, and it is not easy:

I think we need to support Debian stable for just one test build. Mostly because Debian uses old versions of CMake. Maybe it can replace the current min version CMake build running Ubuntu?

Debian Trixie is soon to be released (Aug 2025) and has support for CMake 3.31.6-2 if I get it right.

The latest CMake 4+ does not support min version builds  below 3.5, but I don't think it makes sense to bump to flatcc up to that version and lose OS support.

mikkelfj:
FYI: I have fixed a bunch of CMake warnings that recent CMake versions produce.
At the same time I bumped CMake to 3.25.1 to support Debian 12.
https://github.com/dvidelabs/flatcc/tree/update-cmake
This is work in progress. AppVeyor currently fails due to CMake version.

bjosv:
I added an additional job running Debian 11/Bullseye and 12/Bookworm to the weekly run.

My idea of having the `cmake-minimum-required` job is to be able to test the exact version we have given in the CMakeLists.
We then can make sure that any change in the CMakeLists would work regardless of what version Debian/Ubuntu/xx provides. I believe distributions can lift the CMake version at anytime.

PR #355, opened 2025-08-08
Fix Wimplicit-int-conversion-on-negation

Conarnar:
Clang 21 adds the warning Wimplicit-int-conversion-on-negation which is when a smaller integer type (char, short) is negated and then converted back to to the smaller type, since negation implicitly converts the value to int.

Adding the casts here should stop the warning.


bjosv (approved):

mikkelfj:
It seems this fix has found its way from other commit, so I'll close this one.

PR #363, opened 2025-12-12
Update Windows Build instructions

satur9nine:
Remove `--target` (results in an error)
Add hint to use `camke --help` to determine the correct generator


mikkelfj:
Thanks

PR #365, opened 2025-12-31
Fix compilation when `PPRINTINT_TEST` is set

mbrukman:
* Add missing parens inside `if` statement calling `strcmp()`
* Add missing `0` parameter to `print_bool()`

Before this change:

    $CC pprintint.h                    # success
    $CC -DPPRINTINT_TEST pprintint.h   # failure

After this change:

    $CC pprintint.h                    # success
    $CC -DPPRINTINT_TEST pprintint.h   # success

Tested with both Clang and GCC.


mikkelfj:
thanks

PR #379, opened 2026-04-10
README.md Typo correction and formatting

edlongman:


mikkelfj:
Thanks, I edited a conflict since I just added text below for another change log entry.

PR #383, opened 2026-07-28
Fix broken Weekly CI and modernize runner configuration

bjosv:
- Fix Windows build failure: replace msbuild `cmake --build` to handle VS 2026 on `windows-2025` runner
  Remove action `microsoft/setup-msbuild` since its not needed when we let CMake build.
- Add `Windows Visual Studio 11-arm`
- Replace retired `macos-14` runner with `macos-26`
- Update Debian matrix: drop EOL `Bullseye` (11), add `Trixie` (13)
- Upgrade Ubuntu runner to 26.04 for modern toolchain jobs (keep 24.04/22.04 for legacy compilers)
- Add Clang 22 to weekly matrix
- Bump action versions: checkout v7.0.1, setup-cpp v1.8.1, actions-setup-cmake v2.2.0, run-on-arch-action v3.1.0
  This replaces the dependabot PRs.

The `aminya/setup-cpp` action still gives annotations/warnings when retrying to add apt keys, but it succeeds to install the compilers. (Weekly run: https://github.com/Nordix/flatcc/actions/runs/30353387666)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
 
   macos:
     name: macOS
-    runs-on: macos-14
+    runs-on: macos-26
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - name: Build and run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,12 +49,11 @@ jobs:
 
   windows:
     name: Windows
-    runs-on: windows-2022
+    runs-on: windows-2025
     steps:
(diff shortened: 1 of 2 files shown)


bjosv:
I pushed again, had to revert back to use ubuntu-24.04 in the gcc job, avoiding ubuntu-22.04

bjosv:
I need to fix a flakyness in clang-5 to 15 first, the first attempt did not work. Hold on

mikkelfj:
Impressive regression on LTS ...

bjosv:
Well, now it seems stable, been [rerunning](https://github.com/Nordix/flatcc/actions/runs/30360037542) 5+ times. So these are the proposed updates.

Yeah, LTS sounds good. Then it doesn't help that Github constantly changes what their runners consists of.

mikkelfj:
Awesome, thanks!

PR #391, opened 2026-08-12
Add a way to pad buffer with specified alignment

illiliti:
flatcc_builder_get_direct_buffer_aligned can be used to force alignment for buffer

### DIFF
--- a/src/runtime/builder.c
+++ b/src/runtime/builder.c
@@ -1956,6 +1956,25 @@ void *flatcc_builder_get_direct_buffer(flatcc_builder_t *B, size_t *size_out)
     return 0;
 }
 
+void *flatcc_builder_get_direct_buffer_aligned(flatcc_builder_t *B, size_t *size_out, size_t block_align)
+{
+    if (!B->is_default_emitter) {
+        *size_out = 0;
+        return 0;
+    }
+    size_t size = flatcc_builder_get_buffer_size(B);
+    size_t padding = (-size) & (block_align - 1);
+    if (padding > 0) {
+            iov_state_t iov;
+            init_iov();
+            push_iov(_pad, padding);
+            if (0 == emit_back(B, &iov)) {
+                return 0;
+            }
+    }
+    return flatcc_emitter_get_direct_buffer(&B->default_emit_context, size_out);
+}
+
 void *flatcc_builder_copy_buffer(flatcc_builder_t *B, void *buffer, size_t size)
 {
     /* User is allowed to call tentatively to see if there is support. */


mikkelfj:
A appreciate your efforts working on this so please view this as constructive feedback.

As mentioned in the discussion section, this only address the end padding for direct buffers which only work for small buffers using the default emitter. Thus it does not address the general issue of end padding not working as intended. This should probably be some central internal function that finalizer functions call regardless of how the is retrieved.

The name _aligned in the contect of retrieving finalized buffer refer to the start of the buffer, not the end padding. The fact that aligned buffers also pad alignment at the end (at at least attempts to) is secondary. Furthermore, direct buffers cannot be start aligned:

The default emitter works build builidng a list if fixed sized pages that eventually gets copied into a user provided buffer, or returned via an allocation method, which can aligned or not depending on method. The pages get reused when reusing the same builder.

The direct buffer works when the entire buffer content can fit in the first page and otherwise returns null.

By default vtrables are packed at the end and general content is prepended so the last content added is the first buffer content. In order to deal with that, the builder uses a virtual offset system where vtables start at offset 0 and grows upwards, and regular content prepends at negative offsets.

When the buffer is start aligned, padding is added between the buffer header and the actual content such that if the buffer is copied to an aligned location, all internal offsets will also be aligned to the respective alignment requirements.

Because the direct buffer is returned without copying, the alignment depends on where offset 0 is placed.

Offset 0 is normally placed in the middle of the first page, but it depends on compile time flags. Therefore, normally, if the page is 2K in size, the offset 0 is still at 1 K boundary iff a page starts at a 2K boundary, but it does not, it just regular allocation which is usually 4 or 8 byte aligned depending on platform.

Because you work on end padding, you might not care about these details. But it matters because _aligned means something else than what you think it does. It is essentially pointless to  provide an aligned direct access method.

What we really need to do is :

1. ensure that all final methods the claim to be start aligned, are also end padded to that alignment.
2. Consider if we should end pad all buffers reguardless of how the are retrieved. This makes sense if the impact is small, and would then also work for get direct buffeer without _aligned variant as proposed here.

At the same time, nested buffers significantly complicates the padding start, and we need to also consider padding nested buffers at the end.

Because it is a long time since I wrote the code, I don't recall exactly what block_align does. I think it sets a minimum start alignment, e.g. if you have a GPU or vector instructions that need to process the buffer, but it requires more reading of header comments and code.

illiliti:
Thanks for thoughtful answer! I'll definitely work on a proper solution later, just needed something as a stop gap solution for this padding mess so I can continue working with flatcc.

Also I primarily need this function to pad my flatbuffer messages, so they become a multiple of 8. Yes, exactly 8, not 4 or any other power of 2 because my protocol is structured in such way that messages can be stacked, e.g `<msg1><msg2><msg3>` and so on. And they all sit on the same global buffer ring, so each message must be aligned to 8, otherwise verify will fail, which may result in misaligned access if ignored.

PR #397, opened 2026-09-23
Report an error instead of asserting when the token stream ends early (#386)

### LINKED ISSUE #386: [Bug]reachable assert(0) abort in parser next() when an embedded NUL truncates the token stream (CWE-617)
## Summary

The parser's token-stream `next()` assumes `P->token != P->te` when it advances (`src/compiler/parser.c:357`). A schema containing an embedded NUL byte — here between `root_type` and the type name — truncates the token stream at the lexer level, so when `next()` is asked for the token after `root_type`, the end-of-token sentinel is hit and `assert(0)` fires, aborting the whole compile instead of reporting a clean parse error.

- **Affected versions**: flatcc HEAD `4c3b999e3c0a423f9fd8df3c9e58ba9d2069d71b` (v0.6.4-pre)
- **Severity**: Medium (reachable assertion abort / SIGABRT; crash/DoS of the schema compiler)
- **CWE**: CWE-617 (Reachable Assertion)

### Affected code

```c
/* src/compiler/parser.c:357 (excerpt) */
static void next(flatcc_parser *P)
{
    ...
    assert(P->token != P->te);    /* <-- fires when embedded NUL ends the stream early */
    P->token += P->token_len;
    ...
}
```

**Root cause**: the lexer treats the embedded NUL as end-of-input, so the token queue terminates before the parser has consumed the expected number of tokens. `next()` does not handle the exhausted-token-stream case gracefully — it asserts instead of emitting a syntax error.

**Reachability**: fully attacker-controlled `.fbs` schema text. The NUL byte is legal inside the raw buffer but is never expected inside an identifier context.

Call chain:

```
next (src/compiler/parser.c:357, assert(0))
  └─ __flatcc_fb_parse (src/compiler/parser.c:1489)
       └─ flatcc_parse_buffer (src/compiler/flatcc.c:113)
            └─ LLVMFuzzerTestOneInput (harness.c:66)
```

## POC (tested on the unmodified source)

Attached **`flatcc_poc3_embedded_nul.fbs`** (57 bytes) — a valid schema with a NUL byte (`0x00`) inserted between `root_type` and the type name:

```text
table DoubleVec {
  a: [double];
}

root_type\x00DoubleVec;
```

Raw hex of the trigger region (`root_type` + NUL + `DoubleVec;`):

```
... 726f 6f74 5f74 7970 6500 446f 7562 6c65 5665 633b 0a
```

```bash
./flatcc_schema_parser_fuzzer flatcc_poc3_embedded_nul.fbs
```

## Trigger result

```
Assertion failed: P->token != P->te (src/compiler/parser.c:357)
Aborted (core dumped)
EXIT=134 (SIGABRT)
```

- Replay exit code: `134` (SIGABRT, assert(0))
- Deterministic: yes — the embedded NUL reliably truncates the token stream

## Suggested fix

Replace the `assert(0)` with a proper exhausted-stream error path that reports a syntax error and unwinds cleanly:

```c
if (P->token == P->te) {
    flatcc_parser_error(P, "unexpected end of token stream");
    return;
}
```

Also consider rejecting embedded NUL bytes in the lexer with a diagnostic instead of silently terminating the input.

1820893135-pixel:
`next()` advanced the token pointer and, on reaching the end of the token
stream, hit `assert(0)` **before** the `switch` that was written to handle
exactly that case (`src/compiler/parser.c:357`):

```c
if (P->token == P->te) {
    --P->token;
    assert(0);
    switch (P->token->id) {
    case LEX_TOK_EOS: case LEX_TOK_EOB: case LEX_TOK_EOF:
        P->token->id = LEX_TOK_EOF;
        return P->token;
    }
    error_tok(P, P->token, "unexpected end of input");
}
```

An embedded NUL in the schema truncates the token stream at the lexer level, so
a parser that has not yet consumed everything it expects reaches this branch and
aborts the whole compiler instead of reporting a syntax error.

### Reproducer

57-byte schema (the `\0` is a real NUL byte, `0x00` - not the two characters
`x00`):

```
table DoubleVec {
  a: [double];
}

root_type<0x00>DoubleVec;
```

```
$ xxd schema.fbs | tail -2
00000020: 0a7d 0a0a 726f 6f74 5f74 7970 6500 446f  .}..root_type.Do
00000030: 7562 6c65 5665 633b 0a                   ubleVec;.

$ bin/flatcc_d schema.fbs
flatcc_d: src/compiler/parser.c:357: next: Assertion `0' failed.
Aborted (core dumped)          # rc=134
```

In a release build (`NDEBUG`) the assert is compiled out and the compiler
instead walks into the switch below with the reduced token pointer, so the
debug build is what makes the bug visible.

### Fix

Remove the `assert(0)` so the handling directly below it runs. The same schema
now fails with a normal error and exit status 255:

```
schema.fbs:5:10: error: '': missing ';' expected by root_type at: schema.fbs:5:10: ''
output failed
```

### Testing

- The reproducer above: `SIGABRT` before, clean error after.
- Other NUL positions (`root_type<0x00>T;`, `table T {<0x00>}`, `root_type T<0x00>;`)
  all report errors rather than aborting.
- Schemas without an embedded NUL are unaffected - the 9 `.fbs` files under
  `test/` still compile.

Fixes #386.


mikkelfj:
OK, I assume the assert was a stray debug assert during development.

Can you please remove the comment from the PR, it refers to past history that would make no sense to a new reader, and the code is not special enough to warrant a comment here at all. Note that when injecting a null terminator, this is captured by the token case above the error, allowing the parser to produce a context specific error rather than unexpected end of file.

